PR #6, opened 2021-09-10
restrict test action to PR

mrmasterplan:
I expect that this will restrict the running of the test action to be in the pr, and not after the PR is merged.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,10 +1,6 @@
 name: Test
 
 on:
-  push:
-    branches:
-      - main
-      - master
   pull_request:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -1,10 +1,6 @@
 name: Test
 
 on:
-  push:
-    branches:
-      - main
-      - master
   pull_request:
```
perkops, line 4: suggestion:

on:
  pull_request:
    types:
      - opened
      - synchronize
      - reopened

mrmasterplan: implemented

perkops (approved):

PR #14, opened 2021-09-22
pyspark now not an explicit dependency

mrmasterplan:
pyspark now not an explicit dependency. Installed in the pipelines instead


mrmasterplan:
The version test fails because the last verion conflicts with pyspark. Exactly the problem I need to avoid.

PR #18, opened 2021-09-25
Tuned pipelines

mrmasterplan:
Piplines are renamed and tuned
using concepts from other atc libs:
- pre-integration
- post-integration
- release
except the last two are combined.


.github/workflows/pre-integration.yml
```diff
@@ -0,0 +1,45 @@
+name: Pre-Integarion
```
perkops, line 1: spelling mistake

.github/workflows/post-integration-and-release.yml
```diff
@@ -1,12 +1,12 @@
...
+name: Post-Integration and Release
 
 on:
   push:
     branches:
       - main
 
 jobs:
-  deploy:
+  release:
```
perkops, line 9: Why do we push to both test and "prod" pypi in the same flow - wouldnt this be nice to split up in different flows? E.g. post-integration pushes to test-pypi and release pushes to pypi.

We have the stable step, so we can accumulate changes before we make a release (in order not to spam people).

In next version of this setup, we would want the documentation to be updated.

perkops: Right now, i think it would be best to align the releases processes - until we have time to streamline to new flow/setup.

mrmasterplan: Ok, let's do it your way. post-integration is now equal to publishing to TestPyPi. Release is publishing to PyPi

perkops (approved):

PR #19, opened 2021-09-27
Add custom drop table cascade

LauJohansson:
This PR contains an idea for a function that drops a table and its files. As I understand databricks, there is no way of removing a table including its files. The "cascade" function in SQL only works for a whole database. Let me get your thoughts.


mrmasterplan:
nice stuff. Mind if I touch it up a bit with type-hints?

src/atc/functions.py
```diff
@@ -2,10 +2,47 @@
...
 from pyspark.sql.types import StringType
 import uuid as _uuid
+import pyspark.sql.functions as F
+from pathlib import Path
 
 # Pyspark uuid function implemented as recommended here
 # https://stackoverflow.com/questions/49785108/spark-streaming-with-python-how-to-add-a-uuid-column/50095755
 uuid = udf(lambda: str(_uuid.uuid4()), StringType()).asNondeterministic()  # noqa
+
+def init_dbutils(spark):
```
mrmasterplan, line 16: since we use Spark.get() in drop_table_cascade, is it not ok  to use it here also and remove the argument?

src/atc/functions.py
```diff
@@ -2,10 +2,47 @@
...
+    try:
+        from pyspark.dbutils import DBUtils
+        dbutils = DBUtils(spark)
+    except ImportError:
+        import IPython
+        dbutils = IPython.get_ipython().user_ns["dbutils"]
+    return dbutils
+
+def drop_table_cascade(DataBaseName,TableName):
```
mrmasterplan, line 25: You only use DataBaseName and TableName in the combination `f"{DataBaseName}.{TableName}"`. Is it not ok to just give this as a single argument?

src/atc/functions.py
```diff
@@ -2,10 +2,47 @@
...
+
+    #Check if table exists 
+    if Spark.get()._jsparkSession.catalog().tableExists(DataBaseName, TableName):
+
+        #### GET PATH - Idea 1 ###
```
mrmasterplan, line 30: I definitely prefer your idea 2. Let's remove the discussion comments and just leave the code.

src/atc/functions.py
mrmasterplan, line 25: Also, I recommend to add a comment string that says what the function does.

src/atc/functions.py
```diff
@@ -2,10 +2,47 @@
...
+        # #Collect parent 
+        # table_path=str(Path(file_name).parent)
+        ##########################
+
+        #### GET PATH - Idea 2 ###
+        table_path=str(Spark.get().sql(f"DESCRIBE DETAIL {DataBaseName}.{TableName}").collect()[0]["location"])
+        ##########################
+
+        if table_path is None:
+            raise Exception(f"Table path is NONE.")
```
mrmasterplan, line 42: It is generally a good idea to make up your custom Exception classes and raise those:

    class MyUnableToCascadeException(Exception):
        pass

This allows those that have to catch it to only catch your specific one. In fact pycharm complains if you try to `except Exception` saying that this is too broad (it will catch everything including misspelled variables etc.).

src/atc/functions.py
LauJohansson: Like this?

def init_dbutils():
    try:
        from pyspark.dbutils import DBUtils
        dbutils = DBUtils(Spark.get())
    except ImportError:
        import IPython
        dbutils = IPython.get_ipython().user_ns["dbutils"]
    return dbutils

src/atc/functions.py
LauJohansson: I dont understand how I should do this? Lets have a talk

mrmasterplan (approved):

PR #25, opened 2021-10-04
Fluent interface for the etl framework

mrmasterplan:
I made the fluent interface work that you suggested. I really like the style, plus it was a nice challenge.

TODO: Documentation.

NOTE: This is a pull-request to merge into another feature branch. This is intentional.


christianhelle:
Nice! This is exactly what I had in mind! Thanks!

PR #26, opened 2021-10-04
Feature/spark function to join to time structured dataframes

JeppeBlixen:


mrmasterplan (approved):
This will not break anything existing, so let's merge it in.
I recommend that you increment the version to 0.1.18 (no rc or pre appendix) so that we can get this released.

PR #27, opened 2021-10-04
Version Manipulation

mrmasterplan:
We want effortless continuous release. The price that we are willing to pay is that re relinquish manual control of the micro version. The micro version is now automatically set to the git commit count.

This manipulation happens in the release pipeline and cannot be configured.

Earlier version checks have been removed. If your git hook breaks, please remove the file `.git/hooks/pre-commit`.


utilities/compare_version.py
```diff
@@ -14,29 +14,34 @@
...
 
     version = parse(atc.__version__)
 
-    if not (version > test_pypi_version):
-        print(
-            f"Current version {atc.__version__}"
-            f" is not ahead of TestPyPi's {test_pypi['info']['version']}"
-        )
+    if version in pypi_test:
+        print(f"Current version {atc.__version__} already exists in TestPyPi")
```
LauJohansson, line 23: use the variable "version" which is declared in line 20 ?

LauJohansson: so use "version" variable in stead of atc.__version__

LauJohansson (commented):
I have made two small comments.

utilities/compare_version.py
mrmasterplan: The version variable is a parsed version object which can have a string representation that is different from the actual version.

so as an example `parse('0.1.18pre1') == '0.1.18rc1'`

I therefore do not want to use the version variable.

utilities/compare_version.py
```diff
@@ -14,29 +14,34 @@
...
         return 1
-    if not (version > pypi_version):
-        print(
-            f"Current version {atc.__version__}"
-            f" is not ahead of PyPi's {pypi['info']['version']}"
-        )
+    if version in pypi:
+        print(f"Current version {atc.__version__} already exists in PyPi")
```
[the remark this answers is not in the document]
mrmasterplan: see other comment

christianhelle:
Why not just automate the versioning and re-write `atc.__version__` from the github workflow using the `${{ github.run_number }}`

It should be hard to re-write **setup.cfg** from the github workflow

```
- name: Update Version
  shell: pwsh
  run: |
    $cfg = Get-Content src/setup.cfg
    $cfg[0] = "[metadata]"
    $cfg[1] = "version = 0.1.${{ github.run_number }}"
    $cfg[2] = "license_files = LICENSE"
    Set-Content src/setup.cfg $cfg
```

mrmasterplan:
Call me old fashioned, but all "standard" python repos, like for example [pip](https://github.com/pypa/pip/blob/main/src/pip/__init__.py) have the version in the `__init__.py` file as equivalent to the latest published version. Now, unless we want something in the PR pipeline to take some automatic action to commit an incremented version, people are going to have to remember to increment that version. I don't like automated commits, hence we have this script. I even made a git hook to remind you on every commit.

christianhelle:
So how about

```
- name: Update Version
  working-directory: src/atc
  shell: pwsh
  run: |
    $initFile = '__init__.py'
    $line = Get-Content $initFile | Select-String '__version__' | Select-Object -ExpandProperty Line
    if ($line -ne $null) {
        $content = Get-Content $initFile
        $content | ForEach-Object {$_ -replace $line, '__version__ = "0.1.${{ github.run_number }}"'} | Set-Content $initFile
    }
```

christianhelle:
I think major.minor versions should be manually updated but doesn't the build / revision number usually correspond to the build server on to which the product is built?

In some of my other projects I use major.minor.build-number.git-sha (1.4.1543.07c0da6)

mrmasterplan:
Thanks for the talk, @christianhelle. I have updated the code and pipelines according to our chat.

christianhelle (approved):

PR #38, opened 2021-10-12
Feature/merge dataframe into table

LauJohansson:
Two methods:

merge_df_into_target
and
concat_dfs

### DIFF
--- a/src/atc/transformations.py
+++ b/src/atc/transformations.py
@@ -14,7 +14,7 @@ def join_time_series_dataframes(
     startTimeColumnName: str,
     endTimeColumnName: str,
     idColumns: List[str],
-    stateColumn: str
+    stateColumn: str,
 ) -> DataFrame:
     """
     Join changing states from two dataframes acording to time as follows.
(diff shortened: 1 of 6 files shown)


src/atc/transformations.py
```diff
@@ -197,3 +204,168 @@ def join_time_series_dataframes(
...
+        {', '.join(f"target.{col} = source.{col}" for col in non_join_cols)}
+        WHEN NOT MATCHED
+        THEN INSERT -- insert new records
+        (
+        {', '.join(df.columns)}
+        )
+        VALUES
+        (
+        {', '.join(f"source.{col}" for col in df.columns)}
+        )
```
mrmasterplan, line 315: `WHEN MATCHED THEN UPDATE SET * WHEN NOT MATCHED THEN INSERT *` is much more concise and does the same. See the spark documentation.

src/atc/transformations.py
```diff
@@ -197,3 +204,168 @@ def join_time_series_dataframes(
...
+    all_cols = sorted(list(set(all_cols)))
+
+    # Create a list to save missing columns for each dataframe
+    missing = []
+
+    # For each dataframe: Get missing columns and add them to the dataframe
+    for i, df in enumerate(dfs):
+        missing.append([col for col in all_cols if col not in df.columns])
+
+        for col in missing[i]:
```
mrmasterplan, line 358: - `missing` is never used outside the for loop. there is no reason to keep this state for later.
- the list comprehension that you construct adds nothing here.
- recommendation: do a simple set of nested for loops
```
    for i, df in enumerate(dfs):
         for col in all_cols:
            if col not in df.columns: 
                dfs_updated[i] = dfs_updated[i].withColumn(col, F.lit(None))
```
- this code is much cleaner and would have been easier for me to read.

tests/functions/test_drop_table_cascade.py
```diff
@@ -0,0 +1,9 @@
+import unittest
+
+@unittest.skip("No tests has yet been implemented.")
+# This file should test the function "drop_table_cascade" in atc/functions
```
mrmasterplan, line 4: maybe add a comment that our current test pipeline does not support delta tables yet.
Also in the other skipped test.

LauJohansson: I have rewritten the comment (y).

PR #43, opened 2021-11-01
Feature/add eventhub stream extractor

JeppeBlixen:
## Eventhub stream extractor
This extractor reads data from an Azure eventhub and returns a strctual streaming streaming dataframe.

Under the hood [spark eventhub](https://github.com/Azure/azure-event-hubs-spark) is used.

```python
class EventhubStreamExtractor(Extractor):
    def __init__(self, 
                 consumerGroup: str,
                 connectionString: str = None,
                 namespace: str = None,
                 eventhub: str = None,
                 accessKeyName: str = None,
                 accessKey: str = None,
                 maxEventsPerTrigger: int = 10000):
    ...
```

Usage example with connection string:
``` python
eventhubStreamExtractor = EventhubStreamExtractor(
    consumerGroup="TestConsumerGroup",
    connectionString="TestSecretConnectionString",
    maxEventsPerTrigger = 100000
)
```

Usage example without connection string:
``` python
eventhubStreamExtractor = EventhubStreamExtractor(
    consumerGroup="TestConsumerGroup",
    namespace="TestNamespace",
    eventhub="TestEventhub",
    accessKeyName="TestAccessKeyName",
    accessKey="TestSecretAccessKey",
    maxEventsPerTrigger = 100000
)
```


docs/extractors/README.md
```diff
@@ -0,0 +1,91 @@
+# Extractors documentation
+This page documents all the atc extractors, following the OETL pattern. 
+
+Extractors in atc-dataplatform:
+
+* [Eventhub stream extractor](#eventhub-stream-extractor)
+
+## Eventhub stream extractor
+This extractor reads data from an Azure eventhub and returns a strctual streaming streaming dataframe.
```
LauJohansson, line 9: spelling "structural"

src/atc/extractors/eventhub_stream_extractor.py
```diff
@@ -0,0 +1,40 @@
...
+from atc.spark import Spark
+
+import json
+from pyspark.sql import DataFrame
+
+class EventhubStreamExtractor(Extractor):
+    def __init__(self, consumerGroup: str, connectionString: str = None, namespace: str = None, eventhub: str = None, accessKeyName: str = None, accessKey: str = None, maxEventsPerTrigger: int = 10000):
+        self.spark = Spark.get()
+
+        # If connectionString is missing, create it from namespace, eventhub accessKeyName and accessKey
```
LauJohansson, line 11: What about creating an early exception if one of the inputvalues are none?

src/atc/extractors/eventhub_stream_extractor.py
```diff
@@ -0,0 +1,40 @@
...
+
+        # Define where to start eventhub stream
+        # It can be done from offset, seqence number or timestamp
+        # Below setting will start stream from the beginning
+        startingEventPosition = {
+            "offset": "-1", #Start stream from beginning
+            "seqNo": -1, #not in use
+            "enqueuedTime": None, #not in use
+            "isInclusive": True
+        }
```
LauJohansson, line 23: Should one be able to choose where to start the stream?  If this setting starts the stream from the the beginning by default, could someone else want it to start at another place (than the beginning)?

LauJohansson:
@JeppeBlixen 
Also use the black-formatter for your files. Write e.g. the following in the terminal: black src/atc/

src/atc/extractors/eventhub_stream_extractor.py
JeppeBlixen: That is a good question. I could imagine in most cases the most relevant would be to start from beginning. But maybe also starting from a timestamp.

JeppeBlixen: I have added to possibility to give a start datetime.

LauJohansson (approved):

PR #51, opened 2021-11-21
turns out there was a standard function for this all along

mrmasterplan:
I threw out all my code because I finally found the standard function to do this.


JeppeBlixen (approved):
Looks good.
This is properly not the last time depreciation will be used.

LauJohansson:
This PR has been approved @mrmasterplan :)

PR #56, opened 2022-01-05
Rename MultipleExtractorOrchestrator to MultipleExtractOrchestrator

JeppeBlixen:
Rename MultipleExtractorOrchestrator to MultipleExtractOrchestrator


mrmasterplan (changes requested):
No thanks. We already adapted to the new naming. Don't change it again.

PR #57, opened 2022-01-08
Etl rewritten and simplified

mrmasterplan:
I completely re-thought the whole ETL library. 
**There are backwards incompatible changes** hence the move to version 0.4.
There are no more special cases for multiple vs single datasets on input or output side.
The documentation has been updated to reflect all the simplified changes.
All `Multiple...` and `Delegating...` classes are gone! Have a look at how little code there is left in the `atc.etl` library!
The code can do everything that it could before plus:
- it is now possible to specify the key for the dataset returned by an extractor (for use in multi-transform)
- it is now possible to reference the result of previous extraction steps in case of complex extraction logic (we have seen examples in practice)
- `save_many` now exists although I see no usecase for it.
- All orchestration is now streamlined so any sequence of transform, load, extract steps can be orchestrated easily.
- the examples are now executed as part of unit testing. They can then be inserted into the readme by a simple script.


docs/etl/insert_examples.py
```diff
@@ -0,0 +1,12 @@
+from pathlib import Path
```
christianhelle, line 1: This is such a cool way of building docs!

src/atc/etl/loader.py
```diff
@@ -1,24 +1,31 @@
...
 from pyspark.sql import DataFrame
 
+from atc.etl.types import dataset_group, EtlBase
 
-class Loader:
-    @abstractmethod
-    def save(self, df: DataFrame) -> DataFrame:
-        return df
 
+class Loader(EtlBase):
```
christianhelle, line 8: I agree that a `Loader` is by design a data sink. The `DataFrame` return type was something I added previously for `Loader` implementations that target CosmosDb, where the document ID and partition key are required by the data store but not relevant on the data record. Returning the `DataFrame` with the new columns made sense at the time. In a way it makes more sense that its the `Transformer` that appends these columns but then the `Transformer` is now responsible for data storage specific concerns, and not purely data transformation

src/atc/etl/orchestrator.py
```diff
@@ -1,142 +1,35 @@
...
-class SingleExtractorOrchestrator(Orchestrator):
-    def __init__(self,
-        extractor: Extractor,
-        transformer: Union[Transformer, DelegatingTransformer],
-        loader: Union[Loader, DelegatingLoader]
-    ):
-        self.extractor = extractor
-        self.transformer = transformer
-        self.loader = loader
+    """
```
christianhelle, line 9: I love this!

tests/etl/test_orchestrator.py
```diff
@@ -0,0 +1,16 @@
+import subprocess
+import sys
+import unittest
+from pathlib import Path
+
+
+class OrchestratorTests(unittest.TestCase):
+    def test_examples(self):
```
christianhelle, line 8: How does this work? Will it the test runner understand that the sub-processes are unit tests? and will it propagate individual test results back to the top level process?

christianhelle: I love the fact that this test enforces us to keep the example code alive and up to date

christianhelle (commented):
These changes are really cool!

src/atc/etl/orchestrator.py
```diff
@@ -1,142 +1,35 @@
...
+    It is up to the user of this library that extractors,
```
christianhelle, line 10: I love this!

src/atc/etl/orchestrator.py
christianhelle, line 9: >>    It is up to the user of this library that extractors,
>>   transformers and loaders live up to their names and are not
>>    used in a wrong order.

I love this!

tests/etl/test_orchestrator.py
mrmasterplan: This simply validates that the examples can be executed. Check=True checks that the process exits normally without exception (no error code).

src/atc/etl/loader.py
mrmasterplan: I am familiar with the use case, and it is fine to both add it in the loader or in a transformer. The real reason to not return it here is to make it clear that in te case of loader->loader->loader, the loaders will be called with the same dataframe and not sequentially enriched dataframes.

mrmasterplan:
@JeppeBlixen I would really appreciate your review since I think you use this in production, and some changes are non-compatible. I recently moved to fixed dependencies, so I know my system won't break, but yours might.

examples/etl/basic_example.py
```diff
@@ -2,51 +2,54 @@
...
         df.printSchema()
         return df
 
 
 class NoopLoader(Loader):
-    def save(self, df: DataFrame) -> DataFrame:
-        df.write.format('noop').mode('overwrite').save()
-        return df
+    def save(self, df: DataFrame) -> None:
+        df.write.format("noop").mode("overwrite").save()
```
LauJohansson, line 43: why not returning the df?

examples/etl/multiple_output_example.py
```diff
@@ -2,58 +2,62 @@
...
         df.printSchema()
         return df
 
 
 class NoopSilverLoader(Loader):
-    def save(self, df: DataFrame) -> DataFrame:
-        df.write.format('noop').mode('overwrite').save()
-        return df
+    def save(self, df: DataFrame) -> None:
+        df.write.format("noop").mode("overwrite").save()
```
LauJohansson, line 45: why not returning the df?

examples/etl/multiple_transform_example.py
```diff
@@ -32,18 +37,18 @@ def process(self, df: DataFrame) -> DataFrame:
...
-print('ETL Orchestrator using multiple transformers')
-etl = (Orchestration
-       .extract_from(GuitarExtractor())
-       .transform_with(IntegerColumnTransformer('id'))
-       .transform_with(IntegerColumnTransformer('year'))
-       .load_into(NoopLoader())
-       .build())
+    def save(self, df: DataFrame) -> None:
+        df.write.format("noop").mode("overwrite").save()
+
```
LauJohansson, line 42: why not returning the df?

examples/etl/no_transform_example.py
```diff
@@ -1,37 +1,38 @@
...
+            ),
+        )
 
 
 class NoopLoader(Loader):
-    def save(self, df: DataFrame) -> DataFrame:
-        df.write.format('noop').mode('overwrite').save()
-        return df
+    def save(self, df: DataFrame) -> None:
+        df.write.format("noop").mode("overwrite").save()
```
LauJohansson, line 31: return df?

examples/etl/multiple_input_example.py
```diff
@@ -2,60 +2,70 @@
...
+            ),
+        )
+
+
+class CountryOfOriginTransformer(Transformer):
     def process_many(self, dataset: {}) -> DataFrame:
-        usa_df = dataset['AmericanGuitarExtractor'].withColumn('country', f.lit('USA'))
-        jap_df = dataset['JapaneseGuitarExtractor'].withColumn('country', f.lit('Japan'))
+        usa_df = dataset["AmericanGuitarExtractor"].withColumn("country", f.lit("USA"))
+        jap_df = dataset["japanese"].withColumn("country", f.lit("Japan"))
```
LauJohansson, line 52: Just for extra niceness. Consider renaming the dataset keys to be more similar.

LauJohansson: e.g. "AmericanGuitars" and "JapaneseGuitars"

examples/etl/multiple_input_example.py
```diff
@@ -2,60 +2,70 @@
...
+        df.write.format("noop").mode("overwrite").save()
+
+
+print("ETL Orchestrator using multiple extractors")
+etl = (
+    Orchestrator()
+    .extract_from(AmericanGuitarExtractor())
+    .extract_from(JapaneseGuitarExtractor())
+    .transform_with(CountryOfOriginTransformer())
+    .load_into(NoopLoader())
```
LauJohansson, line 67: great idea to remove the ".build()" method!

examples/etl/multiple_output_example.py
```diff
@@ -2,58 +2,62 @@
...
-
-print('ETL Orchestrator using multiple loaders')
-etl = (Orchestration
-       .extract_from(GuitarExtractor())
-       .transform_with(BasicTransformer())
-       .load_into(NoopSilverLoader())
-       .load_into(NoopGoldLoader())
-       .build())
+    def save(self, df: DataFrame) -> None:
+        df.write.format("noop").mode("overwrite").save()
```
LauJohansson, line 50: return the df?

examples/etl/basic_example.py
mrmasterplan: https://github.com/atc-net/atc-dataplatform/pull/57#discussion_r780823779

examples/etl/multiple_output_example.py
mrmasterplan: https://github.com/atc-net/atc-dataplatform/pull/57#discussion_r780823779

examples/etl/multiple_transform_example.py
mrmasterplan: https://github.com/atc-net/atc-dataplatform/pull/57#discussion_r780823779

examples/etl/multiple_input_example.py
mrmasterplan: I deliberately made them distinctly different to highlight the fact that the key was set manually to a new value.

src/atc/etl/transformer.py
```diff
@@ -1,49 +1,30 @@
...
-        self.inner_transformers = inner_transformers
+    In regards to the etl step, the transformer CONSUMES the inputs
+    and ADDs the result of its transformation stage.
+    """
 
-    def get_transformers(self) -> List[Transformer]:
-        return self.inner_transformers
+    def etl(self, inputs: dataset_group) -> dataset_group:
+        if len(inputs) == 1:
+            for v in inputs.values():
```
JeppeBlixen, line 20: A bit wired to have a for loop for dict with len = 1?
The same could be written this way: next(iter(d.values()))

JeppeBlixen (approved):
I think it looks super great! I have also updated to fixed dependencies, so no worries for merge.

I have added some comments.

src/atc/etl/loader.py
```diff
@@ -1,24 +1,31 @@
...
+    and does not consume or change it.
+    """
 
-class DelegatingLoader(Loader):
-    def __init__(self, inner_loaders: List[Loader]):
-        super().__init__()
-        self.inner_loaders = inner_loaders
+    def etl(self, inputs: dataset_group) -> dataset_group:
+        if len(inputs) == 1:
+            for k, v in inputs.items():
```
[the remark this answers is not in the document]
mrmasterplan: Im not sure what is cleaner, I could also have gone with `list(inputs.values())[0]` I like your suggestion. let's go with that.

src/atc/etl/transformer.py
```diff
@@ -1,49 +1,30 @@
...
-    def process(self, df: DataFrame) -> DataFrame:
-        return df
-
-
-class MultiInputTransformer:
-    @abstractmethod
-    def process_many(self, dataset: Dict[str, DataFrame]) -> DataFrame:
-        pass
 
+class Transformer(EtlBase):
```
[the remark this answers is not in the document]
mrmasterplan: I had considered it. Now I will do it.

christianhelle (approved):

PR #59, opened 2022-01-11
all Etl objects can now exchange arguments

### LINKED ISSUE #58: orchestration parameters
Idea: the orchestrator holds a dict and gives a reference to each step that can be modified and added to.

mrmasterplan:
- all Etl objects can now exchange arguments
- the orchestrator is itself an Etl object, meaning orchestrators can be steps in higher orchestrators
- ToDo: Tests and documentation
- Closes #58

### DIFF
--- a/src/atc/etl/orchestrator.py
+++ b/src/atc/etl/orchestrator.py
@@ -1,23 +1,33 @@
-from typing import List
+from typing import List, Dict, Any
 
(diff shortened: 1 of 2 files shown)


mrmasterplan:
I don't like this style since it creates unnecessary roundabout ways of passing arguments instead of directly in the constructor or another aux method.

PR #60, opened 2022-01-13
composit orchestration

mrmasterplan:
### Principles

All ETL classes, **Orchestrator**, **Extractor**, **Transformer** and **Loader** are ETL objects.
This means that they have a method `etl(self, inputs: dataset_group) -> dataset_group`
(where `dataset_group = Dict[str, DataFrame]`) that transforms a set of import to a set of 
outputs. The special properties of each type are
 - **Extractor** always adds its result to the total set,
 - **Transformer** consumes all inputs and adds a single result dataframe,
 - **Loader** acts as a sink, while passing its input on to the next sink,

The special case of the  **Orchestrator** is that it takes all its steps and executes them
in sequence on its inputs. Running in the default `execute()` method, the inputs are empty,
but an orchestrator can also be added as part of another orchestrator with the `step` method.

For the most general case of a many to many transformation, implement your step by inheriting
from the `EtlBase` class.

### DIFF
--- a/src/atc/etl/extractor.py
+++ b/src/atc/etl/extractor.py
@@ -15,6 +15,7 @@ class Extractor(EtlBase):
     """
 
     def __init__(self, dataset_key: str = None):
+        super().__init__()
         if dataset_key is None:
             dataset_key = type(self).__name__
         self.dataset_key = dataset_key
--- a/src/atc/etl/orchestrator.py
+++ b/src/atc/etl/orchestrator.py
@@ -5,7 +5,7 @@ from pyspark.sql import DataFrame
 from .types import EtlBase, dataset_group
 
 
-class Orchestrator:
+class Orchestrator(EtlBase):
     """
     It is up to the user of this library that extractors,
     transformers and loaders live up to their names and are not
@@ -13,6 +13,7 @@ class Orchestrator:
     """
 
     def __init__(self):
+        super().__init__()
         self.steps: List[EtlBase] = []
 
     def step(self, etl: EtlBase) -> "Orchestrator":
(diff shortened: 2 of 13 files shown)


christianhelle (commented):
Isn't replacing pass-through virtual function implementations with `raise NotImplementedError()` is a breaking change? I'm not actively working with data or this library so I of course don't really mind, but others who depend on this could be affected

JeppeBlixen:
This looks as a great solution, on the issue of having multiple orchestrator layers. I think it would be hreat to have an example og this, both for testing and documentation

mrmasterplan:
I changed this a bit more. `execute` is now an alias for `etl`. 
To really support composit orchestration for now the extractor needs to handle the case that a key was already given. It this is not handled, the transformer may fail. That makes for a terrible error message. I therefore added  some logic to give helpful warnings if you make this exact mistake.

src/atc/etl/orchestrator.py
```diff
@@ -24,7 +25,38 @@ def step(self, etl: EtlBase) -> "Orchestrator":
...
+        inputs = inputs or {}
+
+        # make a shallow copy of the inputs for waring after the first step
+        datasets = inputs.copy()
+        if not self.steps:
+            raise NotImplementedError("The orchestrator has no steps.")
+
+        # treat the fist step differently to warn in case the input was not handled
+        datasets = self.steps[0].etl(datasets)
+        if len(inputs) and (len(inputs) + 1 == len(datasets)):
```
JeppeBlixen, line 38: I do not understand this need to check if the fist step have cleaned up the inputs.
I could imagine a valid scenario where i have an input dataframe, a first step extractor reading a dataframe, and a transformer working on both of them?

src/atc/etl/extractor.py
```diff
@@ -15,6 +15,7 @@ class Extractor(EtlBase):
     """
 
     def __init__(self, dataset_key: str = None):
+        super().__init__()
```
JeppeBlixen, line 18: This is maybe a detail, but all these __init__ updates to call the parent constructors. Should these be in its own PR? It does not feel like its related composit-orchestration.

src/atc/etl/orchestrator.py
mrmasterplan: Absolutely, this is a special case which is why it is only a warning. Do we need a flag to disable the warning?

src/atc/etl/extractor.py
mrmasterplan: they are generally good practice I believe, and do no harm.

src/atc/etl/orchestrator.py
mrmasterplan: I used the warnings package so that it is only raised the first time. In additon, I added a suppression flag.

JeppeBlixen (approved):

PR #61, opened 2022-01-15
creation of test dataframes

mrmasterplan:
### DataframeCreator

The `DataframeCreator` is a helper class to assist in writing concise unittests.

Unittests typically take a dataframe, often created with `spark.createDataFrame` and transform it.
The function `createDataFrame` requires all data fields to be assigned a value, even if the given unittest is only concerned with a small subset of them.

This class allows the user to specify which columns she wants to give values for. All other columns will be assigned *null* values.

#### Usage:

```python3
from atc.utils import DataframeCreator
from pyspark.sql import types

df = DataframeCreator.make_partial(
            schema=types._parse_datatype_string("""
                Id INTEGER,
                measured DOUBLE,
                customer STRUCT<
                    name:STRING,
                    address:STRING
                >,
                product_nos ARRAY<STRUCT<
                    no:INTEGER,
                    name:STRING
                >>
            """),
            columns=[
              "Id", 
              # of the customer structure, only specify the name
              ("customer", ["name"]),
              # of the products array of structures, only specify the 'no' field in each row
              ("product_nos", ["no"])
            ],
            data=[
                (1, ("otto",), [(1,), (2,)]),
                (2, ("max",), []),
            ],
        )
df.show()
```
Result:
```
| Id|measured|    customer|         product_nos|
+---+--------+------------+--------------------+
|  1|    null|{otto, null}|[{1, null}, {2, n...|
|  2|    null| {max, null}|                  []|
+---+--------+------------+--------------------+
```


mrmasterplan:
what about data types that are of type `ARRAY<STRUCT<a:INT,...>>` maybe it should also be possible to specify their types

christianhelle (commented):
As much as I like the code and the quality of the code, I don't think this belongs in this library. 

This library is designed to run in production environments. I can't see any real production use cases for building a data frames from scratch with test data

Wouldn't it be better to create a new project specifically for something like?

mrmasterplan:
@christianhelle while I have sympathy with your concerns in principle, I actually need code like this in my product this week, and hence I would like to merge this PR as is.

I also don’t want to create a separate package in this repo, a package belongs in its own repo. Maybe we should set up `atc-dataplatform-utils` as a repo and as a pip package. When that is up and running we can refactor this stuff and move it over.

Would that work for you as a plan?

christianhelle (approved):

PR #62, opened 2022-01-19
consuming loader

mrmasterplan:
The loader can now consume its inputs. The important side-effect is that it is possible to have multiple load steps that consume different parts of the output.

### Example-7

In some cases a single destination dataset is not sufficient, as seen in this example.
Here we need to use the full power of the EtlBase class to return multiple datasets that
are then loaded by separate loaders.

Note that setting a key in the loader constructor allows it to pick out only one of the
inputs to save.

```
import pyspark.sql.types as t

from atc.etl import Extractor, Loader, Orchestrator, EtlBase
from atc.etl.types import dataset_group
from atc.spark import Spark


class OrdersExtractor(Extractor):
    def __init__(self):
        super().__init__(dataset_key="orders")

    def read(self) -> DataFrame:
        spark = Spark.get()
        return spark.createDataFrame(
            Spark.get().sparkContext.parallelize(
                [
                    (1, "Guitar", 50),
                    (2, "Telescope", 200),
                    (3, "Tablet", 100),
                ]
            ),
            t._parse_datatype_string(
                """
                id INTEGER,
                product STRING,
                price INTEGER
            """
            ),
        )


class PaymentsExtractor(Extractor):
    def __init__(self):
        super().__init__(dataset_key="payments")

    def read(self) -> DataFrame:
        spark = Spark.get()
        return spark.createDataFrame(
            Spark.get().sparkContext.parallelize(
                [
                    (45, 1, 50),
                    (46, 2, 200),
                    (47, 3, 150),
                ]
            ),
            t._parse_datatype_string(
                """
                id INTEGER,
                order_id INTEGER,
                charged_amount INTEGER
            """
            ),
        )


class ReconcilingTransformer(EtlBase):
    def etl(self, inputs: dataset_group) -> dataset_group:
        orders = inputs["orders"]
        payments = inputs["payments"]

        df = orders.join(payments, orders.id == payments.order_id, "left").select(
            orders.id.alias("order_id"),
            payments.id.alias("payment_id"),
            orders.product,
            orders.price,
            payments.charged_amount,
        )

        dispatch = df.filter(df.price == df.charged_amount)
        service_follow_up = df.filter(~(df.price == df.charged_amount))

        return {"dispatch": dispatch, "service_follow_up": service_follow_up}


class DispatchLoader(Loader):
    def __init__(self):
        super().__init__("dispatch")

    def save(self, df: DataFrame) -> None:
        assert df.count() == 2
        print("Orders ready to dispatch:")
        df.show()


class CustomerServiceLoader(Loader):
    def __init__(self):
        super().__init__("service_follow_up")

    def save(self, df: DataFrame) -> None:
        assert df.count() == 1
        print("Orders that need follow-up:")
        df.show()


print("ETL Orchestrator using multiple different loaders")
etl = (
    Orchestrator()
    .extract_from(OrdersExtractor())
    .extract_from(PaymentsExtractor())
    .transform_with(ReconcilingTransformer())
    .load_into(DispatchLoader())
    .load_into(CustomerServiceLoader())
)
result = etl.execute()

```

Result:
```
Orders ready to dispatch:
+--------+----------+---------+-----+--------------+
|order_id|payment_id|  product|price|charged_amount|
+--------+----------+---------+-----+--------------+
|       2|        46|Telescope|  200|           200|
|       1|        45|   Guitar|   50|            50|
+--------+----------+---------+-----+--------------+

Orders that need follow-up:
+--------+----------+-------+-----+--------------+
|order_id|payment_id|product|price|charged_amount|
+--------+----------+-------+-----+--------------+
|       3|        47| Tablet|  100|           150|
+--------+----------+-------+-----+--------------+
```

### DIFF
--- a/src/atc/etl/loader.py
+++ b/src/atc/etl/loader.py
@@ -2,7 +2,8 @@ from abc import abstractmethod
 
 from pyspark.sql import DataFrame
 
-from .types import dataset_group, EtlBase
+from .types import dataset_group
+from .etl import EtlBase
 
 
 class Loader(EtlBase):
@@ -10,22 +11,38 @@ class Loader(EtlBase):
     A loader is a data sink, which is indicated by the fact that
     the save method has no return.
 
-    In regards to the etl step, a loader USES the input dataset(s)
+    In regards to the etl step, a loader can be used in two ways.
+    If no dataset keys are given, it USES the input dataset(s)
     and does not consume or change it.
+    If one or several dataset keys are given, those are CONSUMED
+    and passed on to the save or save_many methods, while the rest
+    are passed on
     """
 
+    def __init__(self, *dataset_keys: str):
+        super().__init__()
+        self.dataset_keys = dataset_keys
+
     def etl(self, inputs: dataset_group) -> dataset_group:
-        if len(inputs) == 1:
-            df = next(iter(inputs.values()))
-            self.save(df)
-            return inputs
-        self.save_many(inputs)
+        if not self.dataset_keys:
+            if len(inputs) == 1:
+                df = next(iter(inputs.values()))
+                self.save(df)
+            else:
+                self.save_many(inputs)
+        else:
+            if len(self.dataset_keys) == 1:
+                self.save(inputs.pop(self.dataset_keys[0]))
+            else:
+                self.save_many({key: inputs[key] for key in self.dataset_keys})
+                for key in self.dataset_keys:
+                    del inputs[key]
         return inputs
 
     @abstractmethod
     def save(self, df: DataFrame) -> None:
-        pass
+        raise NotImplementedError()
 
     @abstractmethod
     def save_many(self, datasets: dataset_group) -> None:
-        pass
+        raise NotImplementedError()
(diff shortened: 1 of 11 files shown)


JeppeBlixen:
This sound very usefull. You actually just got me in advance. In about three/four week we need to implement af solustion where we need to both write some json data, and add an event to an EventHub. The data is linked but not the same. This is the exact solution for this.

mrmasterplan:
@JeppeBlixen can you approve?

src/atc/etl/loader.py
```diff
@@ -2,30 +2,47 @@
...
+                self.save(df)
+            else:
+                self.save_many(inputs)
+        else:
+            if len(self.dataset_keys) == 1:
+                self.save(inputs.pop(self.dataset_keys[0]))
+            else:
+                self.save_many({key: inputs[key] for key in self.dataset_keys})
+                for key in self.dataset_keys:
+                    del inputs[key]
```
JeppeBlixen, line 39: Why are key deleted from the input dict? It could be used by another loader?

JeppeBlixen (approved):

PR #63, opened 2022-01-20
abstract methods raise not implemented error

mrmasterplan:
I once accidentally implemented the wrong abstract method. I implement a `Transformer.transform` and wondered why my transformation was not applied. I think that not implementing one of the abstract methods should raise.

This caused controversy so let's review it as a topic on its own.

Note: Merging into other feature branch with breaking changes.


christianhelle (approved):
Move and fast and break things

PR #64, opened 2022-01-20
no return orchestrator

mrmasterplan:
In some cases I need a multi-return orchestrator. An orchestrator may contain a many-to-many transformation with several `save_many` loaders. The result will be several remaing datasets in the group.
It is not clear which, if any, datasets the orchestrator should return in this case. The logical consequence is that it should return none.

Note: Meging into 0.5 feature branch


mrmasterplan:
Reviewer Note: Start with src/atc/etl/orchestrator.py

The other files just reflect this changed usage of the orchestrator

JeppeBlixen (approved):

PR #66, opened 2022-01-24
some mocking helpers for unittests

mrmasterplan:
No documentation as yet.
I need to mock the extractors and loaders in some cases. This helps me do that.

### DIFF
--- a/src/atc/utils/MockExtractor.py
+++ b/src/atc/utils/MockExtractor.py
@@ -1,0 +1,15 @@
+from unittest.mock import MagicMock
+
+from pyspark.sql import DataFrame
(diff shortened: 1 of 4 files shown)


christianhelle (approved):
Shouldn't this be implemented in https://github.com/atc-net/atc-dataplatform-tools ? I still stand by my views on not having test code running in production systems but you already know that so I'm guessing you actually really need this code right here and now so I approve :)

mrmasterplan:
@christianhelle you are right and I agree. I just need to put in the work to get the other repo running and then I will refactor everything in the `utils` folder

PR #74, opened 2022-03-11
added a storage accound and corresponding secrets

mrmasterplan:


.github/deploy/Utilities/Databricks/New-DatabricksScope.ps1
```diff
@@ -0,0 +1,35 @@
...
+      $keyVaultResourceId
+    )
+
+    $json = Convert-Safe-FromJson -text (databricks secrets list-scopes --output JSON)
+    if($json.scopes){
+      $scopes = $json | Select-Object -expand "scopes" | Select-Object -expand "name"
+      if ($scopes -contains $name) {
+        Write-Host "  The scope '$name' already exists" -ForegroundColor DarkYellow
+      }
+    }
```
LauJohansson, line 25: Is this correct?

If json.scopes has some kind of scope "scopeA". Then the if statement is true. 
The scopes is then extracted to $scopes. BUT, there is no "else"-statement to the "if ($scopes -contains $name)".

What could then happen?
If there should be created "scopeA", and there exists a "scopeB". Then, the algorithm will find that there is scopes, but it does not contain the "scopeA". It should then be created, but it doesnt!

.github/deploy/steps/99-Databricks-Secrets.ps1
```diff
@@ -0,0 +1,21 @@
...
+  $new_keys += $key.name
+  databricks secrets put --scope $db_secrets_scope --key $key.name --string-value $key.key
+  Write-Host "  Added key '$($key.name)'"
+}
+
+$existing_keys = Convert-Safe-FromJson -text (databricks secrets list --scope $db_secrets_scope --output JSON)
+Throw-WhenError -output $existing_keys
+
+foreach($secret in $existing_keys.secrets){
+  if(-not $new_keys.Contains($secret.key)){
```
LauJohansson, line 16: why have this new variable "new_keys"? New_keys is just $keystore.name ? 
This could remove line 7.

LauJohansson (changes requested):

LauJohansson (approved):

PR #77, opened 2022-03-24
table configurator

mrmasterplan:
todo docs. will be handled as Issue


src/atc/config_master/table_configurator.py
```diff
@@ -0,0 +1,205 @@
...
+            if "path" in value:
+                self.table_names[key]["path"] = value["path"].format(
+                    **self.table_arguments
+                )
+
+    def __get_name_entry(self, table_id: str):
+        entry = self.table_names[table_id]
+        while "alias" in entry:  # allow alias of alias
+            entry = self.table_names[entry["alias"]]
+        return entry
```
LauJohansson, line 130: Could this end up in a infinite loop is two entries has an alias to eachtother?

TableNameA:
     alias: TableNameB


TableNameB:
    alias: TableNameA

I hope that no one does this, but should we address this?

PR #87, opened 2022-04-04
eventhub capture reading

mrmasterplan:


src/atc/eh/EventHubCapture.py
```diff
@@ -0,0 +1,210 @@
...
+        partition_path = ""
+
+        # discover each part
+        for c in self.partitioning:
+            # loop over items in path
+            value = None
+            for file_info in dbutils.fs.ls(self.path + "/" + partition_path):
+                # validate
+                if not file_info.path.startswith("dbfs:"):
+                    raise AtcEhLogicException("Unexpected file paths")
```
LauJohansson, line 91: As it only checks if file_info starts with "dbfs:", which will be the first characters in self.path - could this check be moved up to the __init__ method? 

if not (self.path).startswith("dbfs:"):
                    raise AtcEhLogicException("Unexpected file paths")

src/atc/eh/EventHubCapture.py
```diff
@@ -0,0 +1,210 @@
...
+            # we got the partition in the form y=2022/m=04/d=04
+            # extract the parts
+            pat = re.compile("/".join(f"{c}=(?P<{c}>\\d+)" for c in self.partitioning))
+            match = pat.match(max_p)
+            self.max_partition = datetime(
+                year=int(match.group("y")),
+                month=int(match.group("m")),
+                day=int(match.group("d")),
+                hour=int(match.group("h")) if "h" in self.partitioning else 0,
+            )
```
LauJohansson, line 161: when using "datetime" should we be aware of the UTC-problems? (difference between cluster vs. local python time)

src/atc/eh/EventHubCapture.py
```diff
@@ -0,0 +1,210 @@
...
+            additions.append(f"PARTITION ({partition_spec})")
+
+        while additions:
+            # We need to add partitions
+            parts_per_batch = min(len(additions), 24)
+            print(f"adding {parts_per_batch} partitions")
+
+            # add partitions in batch to prevent an sql line with hundreds of parts
+            batch = [additions.pop() for _ in range(parts_per_batch)]
+            Spark.get().sql(f"ALTER TABLE {self.name} ADD " + " ".join(batch))
```
LauJohansson, line 192: Why is the parts_per_batch 24?

src/atc/eh/EventHubCapture.py
```diff
@@ -0,0 +1,210 @@
...
+        self._repair_partitioning()
+        return (
+            Spark.get()
+            .table(self.name)
+            .withColumn(
+                "pdate",
+                f.expr(
+                    "make_timestamp(y,m,d,h,0,0)"
+                    if "h" in self.partitioning
+                    else "make_timestamp(y,m,d,0,0,0)"
```
LauJohansson, line 207: Aware of e.g. which timezone make_timestamp uses

src/atc/eh/EventHubCapture.py
mrmasterplan: for hourly partitions, it adds a day at a time. For if that number is good enough for hourly, why not keep it for the other case.

mrmasterplan:
Depends on #89

src/atc/eh/EventHubCapture.py
mrmasterplan: the rest of the block uses the `.name` part, this is old and I will remove it

src/atc/eh/EventHubCapture.py
mrmasterplan: I have thought about it, but nowhere in this file does it matter what timezone the object is in, so I decided to go with a 'naive' object.

src/atc/eh/EventHubCapture.py
mrmasterplan: good catch.

mrmasterplan:
The pipeline has been tested on the new subscription and works there. Run integration tests pending update of secrets

tests/cluster/eh/test_eh_saving.py
```diff
@@ -1,14 +1,23 @@
...
 from atc.spark import Spark
 
 from .AtcEh import AtcEh
 
 
 class EventHubsTests(unittest.TestCase):
-    def test_publish(self):
+    @classmethod
+    def setUpClass(cls) -> None:
+        TableConfigurator().clear_all_configurations()
```
LauJohansson, line 18: This is not in debug mode - is that on purpose?

mrmasterplan: It makes no difference here since the registered table contains no tags like `{ID}`, so debug and release table are the same.

mrmasterplan:
@JeppeBlixen Do you want to review before we merge? We are missing some good documentation for the class. I hope you can get started with the example in the unittest.

JeppeBlixen (commented):
It looks good to me, some nice features around the partitioning. This could also be used for other time partitioned file sources, for example json. Would it make sense to split this? One part with generalized logic about the partitioning, and another more related to the eventhub with the validation and table creation.

mrmasterplan:
@JeppeBlixen That's great input for the next PR. For now I want to complete this and get on with my vacation. We can talk again after easter.

PR #89, opened 2022-04-10
Mounting spn setup

mrmasterplan:
The pipeline has been extended with a second service principal that is only used for mounting. The mounting service principal is used from the workspace to mount the storage account.

Note: This pipeline will only work once the pipeline secrets have been updated with the service principal with sufficient privileges.


mrmasterplan:
Re-run integration test once the secrets are updated.

LauJohansson (approved):
I think it looks great. I have no comments.

PR #91, opened 2022-04-19
Delta Table access

mrmasterplan:
# DeltaHandle

The `TableConfigurator` contains logic to distinguish between production and 
debug tables. To make full use of this functionality when reading and writing 
delta tables, a convenience class has been provided. Use the class like this

```python
from atc.delta import DeltaHandle, DbHandle

tc = TableConfigurator()
tc.add_resource_path('/my/config/files')

# name is mandatory,
# path is optional
# format is optional. Must equal 'db' if provided
db = DbHandle.from_tc('MyDb')

# quickly create the database
db.create()

# name is mandatory,
# path is optional
# format is optional. Must equal 'delta' if provided
dh = DeltaHandle.from_tc('MyTblId')

# quickly create table without schema
dh.create_hive_table()
df = dh.read()
dh.overwrite(df)
```

This code assumes that there exists a file `/my/config/files/stuff.yml` like:
```yaml
MyDb:
  name: TestDb{ID}
  path: /tmp/testdb{ID}

MyTblId:
  name: TestDb{ID}.testTbl
  path: /tmp/testdb{ID}/testTbl
```

The `{ID}` parts are either replaced with an empty string (production) or with a uuid
if `tc.reset(debug=True)` has been set.


src/atc/delta/delta_handle.py
```diff
@@ -0,0 +1,95 @@
...
+    @classmethod
+    def from_tc(cls, id: str):
+        tc = TableConfigurator()
+        return cls(
+            name=tc.table_name(id),
+            location=tc.table_path(id),
+            data_format=tc.table_property(id, "format", "delta"),
+        )
+
+    def _validate(self):
```
LauJohansson, line 38: """Validates that the name is either `db`.`table` or just `table`.""" (Maybe do this instead of #-comment)

src/atc/delta/delta_handle.py
```diff
@@ -0,0 +1,95 @@
...
+        if self._data_format != "delta":
+            raise DeltaHandleInvalidFormat("Only format delta is supported.")
+
+    def read(self) -> DataFrame:
+        """Read table by path if location is given, otherwise from name."""
+        if self._location:
+            return Spark.get().read.format(self._data_format).load(self._location)
+        return Spark.get().table(self._name)
+
+    def write_or_append(self, df: DataFrame, mode: str, overwriteSchema=None) -> None:
```
LauJohansson, line 60: What is the datatype of "overwriteSchema"?

def write_or_append(self, df: DataFrame, mode: str, overwriteSchema: str ="false") -> None:

writer = df.write.format(self._data_format).mode(mode).option("overwriteSchema", overwriteSchema)

then remove line 62 and 63?

src/atc/delta/delta_handle.py
```diff
@@ -0,0 +1,95 @@
...
+        writer = df.write.format(self._data_format).mode(mode)
+        if overwriteSchema is not None:
+            writer = writer.option("overwriteSchema", overwriteSchema)
+
+        if self._location:
+            return writer.save(self._location)
+
+        return writer.saveAsTable(self._name)
+
+    def overwrite(self, df: DataFrame, overwriteSchema=None) -> None:
```
LauJohansson, line 70: overwriteschema datatype?

src/atc/delta/delta_handle.py
```diff
@@ -0,0 +1,95 @@
...
+        return self.write_or_append(df, "overwrite", overwriteSchema=overwriteSchema)
+
+    def append(self, df: DataFrame, overwriteSchema=None) -> None:
```
LauJohansson, line 73: overwriteschema datatype?

src/atc/delta/delta_handle.py
```diff
@@ -0,0 +1,95 @@
...
+    def truncate(self) -> None:
+        Spark.get().sql(f"TRUNCATE TABLE {self._name};")
+
+    def drop(self) -> None:
+        Spark.get().sql(f"DROP TABLE IF EXISTS {self._name};")
+
+    def drop_and_delete(self) -> None:
+        self.drop()
+        if self._location:
+            init_dbutils().fs.rm(self._location, True)
```
LauJohansson, line 85: This is might not a realistic case. But what if the table actually has a location when it was created - but has not self._location?

We have a function called "drop_table_cascade" that finds the location of the table, removes the files, and then drops the table. It is not necessary that it should be used in this context, but just to remind that the function exists. 
https://github.com/atc-net/atc-dataplatform/blob/main/src/atc/functions.py

src/atc/functions.py
```diff
@@ -27,6 +28,7 @@ def init_dbutils():
     return dbutils
 
 
+@deprecated(details="Use DeltaHandle.drop_and_delete")
```
LauJohansson, line 31: This function is good if you dont know the exact path to the files on which the table is created. 
The function could might still be usefull for some.

tests/cluster/delta/test_delta_class.py
```diff
@@ -0,0 +1,67 @@
...
+                "name": "TestDb{ID}.TestTbl",
+                "path": "/mnt/atc/silver/testdb{ID}/testtbl",
+            },
+        )
+
+        # test instantiation without error
+        DbHandle.from_tc("MyDb")
+        DeltaHandle.from_tc("MyTbl")
+
+    def test_02_write(self):
```
LauJohansson, line 33: Do you want to test, that overwrite of schema actually works? 

Idea:

First write some data with:
df = Spark.get().createDataFrame([(1, "a"), (2, "b")], "id int, name string")
dh.overwrite(df, overwriteSchema)
assert that data is as expected

then use overwrite:
df = Spark.get().createDataFrame([(1, "a","yes"), (2, "b", "no")], "id int, name string, response string").
dh.overwrite(df, overwriteSchema)
assert that data is as expected

tests/cluster/delta/test_delta_class.py
```diff
@@ -0,0 +1,67 @@
...
+        dh.truncate()
+        df = dh.read()
+        self.assertEqual(0, df.count())
+
+    def test_06_delete(self):
+        dh = DeltaHandle.from_tc("MyTbl")
+        dh.drop_and_delete()
+
+        with self.assertRaises(AnalysisException):
+            dh.read()
```
LauJohansson, line 67: assert that no data in located at "self._location" 
fs.rm(self._location)

tests/cluster/functions/test_drop_table_cascade.py
```diff
@@ -1,81 +0,0 @@
-import unittest
```
LauJohansson, line 1: should we still have the possibility for others to use this function?
It is not a function that exists in databricks, so maybe we could still have it even though we dont use it 1:1 in the TableConfigurator-setup.

LauJohansson (changes requested):

src/atc/delta/delta_handle.py
mrmasterplan: i want this to be a tri-state bool True/False/None. Because any option set here overrides the cluster-wide settings which may not be what you want. Therefore if the value is None, I do not apply it.

src/atc/delta/delta_handle.py
mrmasterplan: the other function is only marked as deprecated, not removed.

src/atc/functions.py
mrmasterplan: only deprecated.

tests/cluster/delta/test_delta_class.py
mrmasterplan: agreed.

tests/cluster/functions/test_drop_table_cascade.py
mrmasterplan: only tests were removed

tests/cluster/delta/test_delta_class.py
mrmasterplan: there will not be data which is which the Analysis Exception is raised. Otherwise it would just be  a zero rows frame.

LauJohansson (approved):

PR #93, opened 2022-05-03
autocreate and read-only mount

mrmasterplan:
An extensive fix to the eventhub class.
- The previous implementation went into an infinite loop.
- reading from a read-only mount did not work
- auto-create can now be disabled.


mrmasterplan:
Integration tests ran fine. This bug-fix needs no review.

PR #100, opened 2022-05-23
cache loader

mrmasterplan:
## Cache Loader

The Cache Loader is provided as `atc.cache.CacheLoader`. The cached loader 
passes only some rows on to the write operation, and not all. This can give 
optimizations for cases where the write operation is costly, such as writes 
to eventhub or cosmos. 

The following diagram illustrates the flow of data 
through the cached loader.
Cache loader diagram

The foundation of the cached loader is a cache table that contains primary key 
columns and a hash for each row. 
The cached loader supports the following use-cases:
- write rows that are 
  - new wrt. the cache table
  - changed wrt. the cache table
- delete rows that are in the cache table but not in the data.

Use the class `atc.cache.CacheLoaderParameters` to configure all these features.
See the unittests for a full example for how to use the CachedLoader.

As a user of this base-class, inherit from it and override the methods 
`write_operation` and `delete_operation`. Only those rows that are returned
by your operation are merged back into the cache. You are therefore free
to apply further filters and limits at this stage.

### `write_operation`

The function receives a dataframe with those payload rows that the caching
logic has determined need to be written. In some cases, the user may want to
add further restrictions, such as limiting the size of written data.
Also, the user may want to add columns that identify the write operation in
other ways than simply by the writing timestamp (which the CachedLoader adds 
automatically). In this case, use the `cache_id_cols` parameters to call 
out these columns. These columns will be added to the cache table. This enables
you to find a given written batch and, for example, invalidate the cache so 
these rows are written again.

Remember that the caching logic makes a hash of all columns in the payload 
rows. Should you wish to add columns that are variable, such as adding a 
"publishedTimestamp", they should be added to the dataframe in the 
`write_operation` method, otherwise the row would always disagree with its
cached version and be written on every run.

Finally, whether you use the `cache_id_cols` or not, you need to return the 
rows that were actually written. The cache loader takes the returned rows and
updates the cache table with them, to mark them as having been written.

### `delete_operation`

The function receives a dataframe with those payload rows that the caching
logic has determined need to be removed, since they are missing in the input 
data. In some cases, the user may want to add further restrictions. In other
cases the user may not want to use the deleting functionality at all.
Therefore, this method needs to return the rows that were actually deleted,
(or return `None`). The cache loader takes the returned rows and
updates the cache table with them, to mark them as having been deleted.
returning `None` here skips the rest of the deleting logic.


.github/deploy/db_token.ps1
```diff
@@ -0,0 +1,15 @@
...
+  --resource-group $resourceGroupName `
+  --name $databricksName `
+  --resource-type "Microsoft.Databricks/workspaces" `
+  --query properties.workspaceUrl `
+  --out tsv
+
+atc_az_databricks_token `
+  --tenantId $tenantId `
+  --appId $authAppId`
+  --workspaceUrl $workspaceUrl
```
LauJohansson, line 15: Does this have something to do with the cache loader? 
You could consider splitting up the PR into multiple PRs which each implements a new feature. This both lets us track the changes, and makes it easier to find out what new features might cause challenges.

docs/etl/specializations/README.md
```diff
@@ -0,0 +1,31 @@
...
+## Cache Loader
+
+The Cache Loader is provided as `atc.cache.CacheLoader`. The cached loader 
+passes only some rows on to the write operation, and not all. This can give 
+optimizations for cases where the write operation is costly, such as writes 
+to eventhub or cosmos. 
+
+The following diagram illustrates the flow of data 
+through the cached loader.
+![Cache loader diagram](./cached_loader.svg)
```
LauJohansson, line 15: You could give the image a bagground colour, since it is very hard to read the image when the GitHub is in dark-mode.

docs/etl/specializations/cached_loader.svg
```diff
@@ -0,0 +1,4 @@
...
....5)rotate(-90 412.5 644.5)"><switch><foreignObject pointer-events="none" width="100%" height="100%" requiredFeatures="http://www.w3.org/TR/SVG11/feature#Extensibility" style="overflow: visible; text-align: left;"><div xmlns="http://www.w3.org/1999/xhtml" style="display: flex; align-items: unsafe center; justify-content: unsafe flex-start; width: 183px; height: 1px; padding-top: 645px; margin-left: 413px;"><div data-drawio-colors="color: rgb(0, 0, 0); " style="box-sizing: border-box; font-size: 0px; text-align: left;"><div style="display: inline-block; font-size: 12px; font-family: Helvetica; color: rgb(0, 0, 0); line-height: 1.2; pointer-events: all; white-space: normal; overflow-wrap: normal;">(return data frame)</div></div></div></foreignObject><text x="413" y="648" fill="rgb(0, 0, 0)" font-family="Helvetica" font-size="12px">(return data frame)</text></switch></g></g><switch><g requiredFeatures="http://www.w3.org/TR/SVG11/feature#Extensibility"/><a transform="translate(0,-5)" xlink:href="https://www.diagrams.net/doc/faq/svg-export-text-problems" target="_blank"><text text-anchor="middle" font-size="10px" x="50%" y="100%">Text is not SVG - cannot display</text></a></switch></svg>
```
LauJohansson, line 4: Consider write "hash" instead of "#" in the cache-table symbol.

src/atc/cache/CachedLoader.py
```diff
@@ -0,0 +1,215 @@
...
+from atc.etl import Loader
+from atc.spark import Spark
+
+from .CachedLoaderParameters import CachedLoaderParameters
+
+
+class CachedLoader(Loader):
+    """
+    Remember to override self.write_operation and/or self.delete_operation
+    Any variable rows need to be added in this function.
```
LauJohansson, line 17: Make a short description of the function here.

src/atc/cache/CachedLoaderParameters.py
```diff
@@ -0,0 +1,29 @@
...
+    ):
+        """
+        Args:
+            cache_table_name: The table that holds the cache
+            key_cols: the set of columns that form the primary key for a row
+            cache_id_cols: These columns, added by the write operation, will be saved
+                in the cache to identify e.g. the written batch.
+        """
+
+        self.cache_id_cols = cache_id_cols or []
```
LauJohansson, line 21: These cache_id_cols, are those just optionally columns in the cache_table? Are they used for anything other than just saved in the table?

src/atc/cache/CachedLoaderParameters.py
```diff
@@ -0,0 +1,29 @@
...
+        self.cache_table_name = cache_table_name
```
LauJohansson, line 22: if not cache_table_name:
            raise ValueError("cache_table_namecolumns must be provided")

src/atc/cache/CachedLoader.py
```diff
@@ -0,0 +1,215 @@
...
+            raise AssertionError(
+                "overload of write_operation " "or delete_operation required"
+            )
+
+        if self.__class__ is CachedLoader:
+            raise AssertionError("You should inherit from this class")
+
+    def write_operation(self, df: DataFrame) -> Optional[DataFrame]:
+        """Abstract Method to be overridden in child."""
+        return None
```
LauJohansson, line 60: Use: raise NotImplementedError

src/atc/cache/CachedLoader.py
```diff
@@ -0,0 +1,215 @@
...
+
+    def delete_operation(self, df: DataFrame) -> Optional[DataFrame]:
+        """Abstract Method to be overridden in child."""
+        return None
```
LauJohansson, line 64: Use: raise NotImplementedError

src/atc/cache/CachedLoader.py
```diff
@@ -0,0 +1,215 @@
...
+            f.current_timestamp().alias(self.params.deletedTime),
+            *self.params.cache_id_cols,
+        )
+
+    def _load_cache(self, cache_to_load: DataFrame) -> None:
+
+        view_name = (
+            self.params.cache_table_name.replace(".", "_")
+            + f"_Update{randint(0, 1000):03}"
+        )
```
LauJohansson, line 143: In atc.function there is a "get_unique_tempview_name()" you could consider to use.

docs/etl/specializations/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+The foundation of the cached loader is a cache table that contains primary key 
+columns and a hash for each row. 
+The cached loader supports the following use-cases:
+- write rows that are 
+  - new wrt. the cache table
+  - changed wrt. the cache table
+- delete rows that are in the cache table but not in the data.
```
LauJohansson, line 23: So the data from the dataplatform (databricks) always makes a Full load when using this class? As I understand it, if you only selected a subset of a databricks gold table, and uses the cache loader, all data (except the subset data) will be deleted - right?

src/atc/cache/CachedLoader.py
```diff
@@ -0,0 +1,215 @@
...
+    def save(self, df: DataFrame) -> None:
+        in_cols = df.columns
+
+        if not set(self.params.key_cols).issubset(in_cols):
+            raise AssertionError(
+                "The key columns must be given in the input dataframe."
+            )
+
+        # ensure that no duplicates exist. The keys have to be a unique set.
+        df = df.dropDuplicates(self.params.key_cols)
```
LauJohansson, line 75: dropDuplicates is not always consistent in the way, that the duplicate chosen can vary.

LauJohansson (changes requested):

.github/deploy/db_token.ps1
mrmasterplan: fair point. i took it out

docs/etl/specializations/README.md
mrmasterplan: Only if you implement the delete functionality, For this, let the `delete_operation` return a dataframe. If `delete_operation` returns None, no deletes will be carried out and no deletes will be registered.

src/atc/cache/CachedLoader.py
mrmasterplan: Indeed, and you should always make sure your primary keys are unique before you use the CachedLoader. This is just an insurance against corrupting the tables.

src/atc/cache/CachedLoader.py
mrmasterplan: imlemented.

src/atc/cache/CachedLoader.py
mrmasterplan: implemented

src/atc/cache/CachedLoaderParameters.py
mrmasterplan: I will add documentation

src/atc/cache/CachedLoader.py
mrmasterplan: implemented

docs/etl/specializations/cached_loader.svg
mrmasterplan: implemented

docs/etl/specializations/README.md
mrmasterplan: done

src/atc/cache/CachedLoaderParameters.py
mrmasterplan: this is alreadu checked as part of the `validate` method

LauJohansson (approved):

PR #102, opened 2022-06-03
dbx update fix

mrmasterplan:
With the latest update [dbx now supports templates in subfolders](https://github.com/databrickslabs/dbx/blob/main/CHANGELOG.md#050---2022-06-01).

At the same time, automatically importing the template from the same folder as the file now fails. Hence we had to update.


mrmasterplan:
pipline failed in delete step. not important. merging anyway.

PR #111, opened 2022-08-02
simple sql server loader and transformer

LauJohansson:


LauJohansson:
@mrmasterplan - i need your skills to allign with the SimpleLoader class :)

mrmasterplan (approved):
I really like it. Approved as-is.

PR #113, opened 2022-08-03
use python formatting in examples

LauJohansson:
Some of the ETL examples needs python formatting so it is easier to read.


LauJohansson:
Since this is only changes in the docs - I skip the integration test.

PR #117, opened 2022-08-03
trunc second when timestamp

LauJohansson:


mrmasterplan:
This is a non-obvious change. Please add a unittest to show how this works.

mrmasterplan (changes requested):
please add a unittest for this special case

LauJohansson:
> please add a unittest for this special case

@mrmasterplan , I have now provided a testcase, where the datetime column had been date_trunc(). It tests that it only has seconds, and no miliseconds.

mrmasterplan (approved):

PR #121, opened 2022-08-08
add exclude pattern

mrmasterplan:


tests/cluster/delta/test_sparkexecutor.py
```diff
@@ -33,8 +33,10 @@ def tearDownClass(cls):
         cls.dh.from_tc("SparkTestTable1").drop()
 
     def test_can_execute(self):
-        SparkSqlExecutor().execute_sql_file("test1")
+        SparkSqlExecutor().execute_sql_file("*", exclude_pattern="debug")
 
         self.dh.from_tc("SparkTestTable1").read()
```
LauJohansson, line 38: To ensure that this actually tests if exclude works, the test should check that neither db2 or table2 exists here.

LauJohansson (changes requested):

tests/cluster/delta/test_sparkexecutor.py
LauJohansson: table1_name = self.tc.table_name("SqlTestTable1")
sql_argument = f"""
        (SELECT * FROM INFORMATION_SCHEMA.TABLES
        WHERE TABLE_NAME = '{table1_name}') target
        """
table_exists = self.sql_server.load_sql(sql_argument)
self.assertEqual(table_exists.count(), 0)

LauJohansson: The code is taken/inspired by the deliverysql-test in atc.

mrmasterplan: thanks for the suggestion. I added this line
```
self.assertEqual(Spark.get().sql("SHOW DATABASES LIKE my_db2*").count(), 0)
```

PR #125, opened 2022-08-12
union transformer

mrmasterplan:
This transformer is so simple that it does not need a unittest

### DIFF
--- a/src/atc/transformers/union_transformer.py
+++ b/src/atc/transformers/union_transformer.py
@@ -1,0 +1,18 @@
+from pyspark.sql import DataFrame
+
+from atc.etl import Transformer
+from atc.etl.types import dataset_group
+
+
+class UnionTransformer(Transformer):
+    """Returns the union of all input dataframes."""
+
+    def process(self, df: DataFrame) -> DataFrame:
+        return df
+
+    def process_many(self, datasets: dataset_group) -> DataFrame:
+        dfs = list(datasets.values())
+        df = dfs[0]
+        for other in dfs[1:]:
+            df = df.union(other)
+        return df
--- a/src/atc/transformers/union_transformer.py
+++ b/src/atc/transformers/union_transformer.py
@@ -7,6 +7,9 @@ from atc.etl.types import dataset_group
 class UnionTransformer(Transformer):
     """Returns the union of all input dataframes."""
 
+    def __init__(self, allowMissingColumns=False):
+        self.allowMissingColumns = allowMissingColumns
+
     def process(self, df: DataFrame) -> DataFrame:
         return df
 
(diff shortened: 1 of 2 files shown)


src/atc/transformers/union_transformer.py
```diff
@@ -0,0 +1,18 @@
...
+    """Returns the union of all input dataframes."""
+
+    def process(self, df: DataFrame) -> DataFrame:
+        return df
+
+    def process_many(self, datasets: dataset_group) -> DataFrame:
+        dfs = list(datasets.values())
+        df = dfs[0]
+        for other in dfs[1:]:
+            df = df.union(other)
```
LauJohansson, line 17: Please keep in mind the difference between union and unionByName. https://spark.apache.org/docs/3.1.1/api/python/reference/api/pyspark.sql.DataFrame.unionByName.html

LauJohansson: also remember that if the dataframes has different columns, you can set allowMissingColumns=True. Again, see documentation.

src/atc/transformers/union_transformer.py
```diff
@@ -0,0 +1,18 @@
...
+class UnionTransformer(Transformer):
+    """Returns the union of all input dataframes."""
+
+    def process(self, df: DataFrame) -> DataFrame:
+        return df
+
+    def process_many(self, datasets: dataset_group) -> DataFrame:
+        dfs = list(datasets.values())
+        df = dfs[0]
+        for other in dfs[1:]:
```
LauJohansson, line 16: what happens if the datasets only contain one dataset?

src/atc/transformers/union_transformer.py
```diff
@@ -0,0 +1,18 @@
...
+from atc.etl.types import dataset_group
+
+
+class UnionTransformer(Transformer):
+    """Returns the union of all input dataframes."""
+
+    def process(self, df: DataFrame) -> DataFrame:
+        return df
+
+    def process_many(self, datasets: dataset_group) -> DataFrame:
```
LauJohansson, line 13: what if something earlier in the ETL has made one of the dataset = NONE. Then it cannot union. Consider something like this:
```
for df in datasets:
        if df.values() is None:
            raise NoTableException(f"The {df.key} table dataframe in the list is None.")
```

LauJohansson:
> This transformer is so simple that it does not need a unittest

Perhaps it does need one?

What do you expect if there is only one dataframe?
What do you expect if the dataframes have different schemas?
What if a dataframe is NONE?

LauJohansson (changes requested):

mrmasterplan:
> > This transformer is so simple that it does not need a unittest
> 
> Perhaps it does need one?
> 
> What do you expect if there is only one dataframe? What do you expect if the dataframes have different schemas? What if a dataframe is NONE?

this is basically a wrapper to the union function. The answer to all those questions is basically whatever the union function does.

LauJohansson (approved):

Issue #123, opened 2022-08-09
yml name should be empty?


LauJohansson:
What you dont want to have a name. See also issue #118

Issue #118, opened 2022-08-05
custom yml format?


LauJohansson:
What if someone want to create another field in the .yml files?

Is this possible? SHOULD it even be possible? 

Something:
      name:
      partitioning:
      customfield:

See also issue #123

PR #126, opened 2022-08-12
table configurator can take any key

mrmasterplan:
TableConfigurator no longer checks the shape of objects. Any keys can now be used. Even name and path are now optional.


RadekBuczkowski (approved):

mrmasterplan:
Merge without integration testing. This PR only affects the locally tested parts

LauJohansson:
Does this solve #118  and #123  ?

Issue #124, opened 2022-08-11
sql handler


LauJohansson:
Create as delta handle class

PR #129, opened 2022-08-15
sqlhandler

LauJohansson:
This PR introduces a sql which should correspond to the deltahandler class (just for azure sql).


mrmasterplan:
closes #124

PR #131, opened 2022-08-19
reading from uninitialized capture now works

mrmasterplan:


LauJohansson:
@mrmasterplan - is this still relevant?

mrmasterplan:
I forgot about this. The code works and contributes value, so let's use it, even though the class is deprecated.

PR #140, opened 2022-08-31
create upsert for deltahandle

LauJohansson:


LauJohansson:
I will wait for #143 before merging.... this PR and 143 needs to be alligned regarding ThHandler.

src/atc/delta/delta_handle.py
```diff
@@ -99,6 +103,88 @@ def recreate_hive_table(self):
...
-        # This should be a method which can merge from delta to delta
-        raise NotImplementedError
+    def get_tablename(self) -> str:
+        return self._name
+
+    def upsert(
+        self,
+        df: DataFrame,
+        join_cols: List[str],
+        incremental_load: bool = True,
```
JeppeBlixen, line 113: Why have the parameter incremental_load? As i se it, incremental_load=False result in a write_or_append(mode="overwrite"), so why not just call that from the user.

src/atc/delta/delta_handle.py
```diff
@@ -99,6 +103,88 @@ def recreate_hive_table(self):
...
+                else "Incremental Base - full load with append"
+            )
+            return df_source
+
+        if df is None:
+            return None
+
+        # check null keys in our dataframe.
+        any_null_keys = len(
+            df.filter(" OR ".join(f"({col} is NULL)" for col in join_cols)).take(1)
```
JeppeBlixen, line 134: If len of this filter is more than 0, the filter is applied, so why not just run the filter on data directly?

src/atc/delta/delta_handle.py
```diff
@@ -99,6 +103,88 @@ def recreate_hive_table(self):
...
+            df = df.filter(" AND ".join(f"({col} is NOT NULL)" for col in join_cols))
+
+        # Load data from the target table for the purpose of incremental load
+        if not incremental_load:
+            return do_simple_save(df, incremental_load_local=False)
+
+        df_target = self.read()
+
+        # If the target is empty, always do faster full load
+        if len(df_target.take(1)) == 0:
```
JeppeBlixen, line 151: This staement of full load when target is empty. Could be hanndled by the merge_required.

src/atc/delta/delta_handle.py
```diff
@@ -99,6 +103,88 @@ def recreate_hive_table(self):
...
+    def get_tablename(self) -> str:
+        return self._name
+
+    def upsert(
+        self,
+        df: DataFrame,
+        join_cols: List[str],
+        incremental_load: bool = True,
+    ) -> Union[DataFrame, None]:
+        def do_simple_save(
```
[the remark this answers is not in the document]
LauJohansson: You are absolutely right! I will change that immediately.

src/atc/delta/delta_handle.py
LauJohansson: Agree. If one needs to make fullload, one should use write_or_append(mode="overwrite").

src/atc/delta/delta_handle.py
LauJohansson: I have removed this check. And discards rows with null in join cols.

src/atc/delta/delta_handle.py
LauJohansson: It could. But this will force an overwrite, where CheckDfMerge forces an append. I will keep them splittet for avoiding too much complexity.

PR #141, opened 2022-08-31
Bicep deployment

LauJohansson:
To utilize the many benefits of Bicep, I will refactor most of our resource deployment to use Bicep. Some of our deployment scripts uses custom powershell code regarding Databricks setup. Therefore, I have combined powershell scripts and Bicep. Since the configurations should be parsed to both it is located in the 00-Config.ps1 file, and send as parameters to the main.bicep file from deploy.ps1.


TomMalow (approved):

christianhelle (approved):
This looks great!

PR #143, opened 2022-09-01
Add upsert logic for sql server

JeppeBlixen:


JeppeBlixen:
First step to create common sql and delta merge query

LauJohansson:
@JeppeBlixen - we use black, isort and flake8. You can wither add them to pre-commits or use File Watchers in your IDE.

src/atc/etl/loaders/UpsertLoader.py
```diff
@@ -51,6 +51,9 @@ def save(self, df_source: DataFrame):
             df = df.filter(
                 " AND ".join(f"({col} is NOT NULL)" for col in self.params.join_cols)
             )
+            df = df.filter(
+                " AND ".join(f"({col} is NOT NULL)" for col in self.params.join_cols)
+            )
```
LauJohansson, line 56: This seems redundant...

src/atc/sql/SqlServer.py
```diff
@@ -100,6 +104,48 @@ def write_table_by_name(
...
+            source_table_name=staging_table_name,
+            join_cols=join_cols,
+            insert_cols=df_source.columns,
+            update_cols=df_source.columns,
+        )
+
+        # Merge query is combined with drop staging table
+        # to ensure the drop of the staging table in case
+        # the problem stops in the midle of the upsert
+        mergeQuery += f" DROP TABLE {staging_table_name};"
```
LauJohansson, line 145: should the mergestatement + drop table be in a transation statement? https://docs.microsoft.com/en-us/sql/t-sql/language-elements/begin-transaction-transact-sql?view=sql-server-ver16

LauJohansson (changes requested):

src/atc/sql/SqlServer.py
JeppeBlixen: I want to be sure that the table is dropped. I have seen examples where the job is stoppd in the middle of the merge(ex when stopping a stream). In this case the case the staging tables would pile up, as this happens.

LauJohansson (approved):

PR #145, opened 2022-09-02
Feature/select-and-drop-transformers

AndreasLindhardt:


src/atc/etl/extended_transformer.py
```diff
@@ -0,0 +1,57 @@
...
+                    if datasetKey in self.dataset_input_key_list
+                }
+            )
+        elif len(inputs) == 1:
+            df = self.process(next(iter(inputs.values())))
+        else:
+            df = self.process_many(inputs)
+
+        inputs[self.dataset_output_key] = df
+        return inputs
```
LauJohansson, line 49: I would really like to have a unit-test of these logics, if this extended transformer is going to be used as a core class to ETL processes. There is some logic (if-else statements), and I think it is neccesary to tests that it behaves as expected. You should might consult with @JeppeBlixen

LauJohansson (changes requested):

src/atc/etl/extended_transformer.py
LauJohansson: Also, if this is a kind of core transformer class, it should also get some documentation on what it is, and how to use it.

LauJohansson: I have consulted with Jeppe, and @MartinBoge is going to implement this incl. with tests. I would recommend that you wait for that PR, merge it into this PR, and then remove the "ExtendedTransformer" and let DropColumnsTransformer and SelectColumnsTransformer inherit Martins transformer.

mrmasterplan:
what does the extended consumer do that the standard comsumer does not do? Please explain. I currently cannot see the need to add this core class.

JeppeBlixen:
@mrmasterplan and @LauJohansson. This pr was just to add out generic drop and select transformers. Could we move this discussion about the ExtendedTransformer to this pr: https://github.com/atc-net/atc-dataplatform/pull/149 Martin is working on this part.

LauJohansson:
> @mrmasterplan and @LauJohansson. This pr was just to add out generic drop and select transformers. Could we move this discussion about the ExtendedTransformer to this pr: #149 Martin is working on this part.

Yes!

LauJohansson:
I have changed the name of the PR to a more suitable name.

PR #149, opened 2022-09-07
Feature/extended transformer 2.0

MartinBoge:
This PR introduces the ExtendedTransformer and ExtendedLoader core classes. 

The ExtendedTransformer class will not consume the input dataset. A transformer, which inherits from this extended class, will therefore append to the existing dataset of keys and dataframes. This will allow for greater flexibility in etl flows, thus allowing the user to reuse dataframes, by not "loosing" them to a transformer. The ExtendedTransformer class uses input- and output key(s) to ensure transformation of the desired dataframes as well as allowing the user to have complete control of the output frame. The ExtendedTransformer returns a single dataframe. 

The ExtendedLoader class has the same functionality as the core Loader class, but it will also handle input key(s). It is crucial to be able to pass these keys in the load (etl step), such that the ExtendedLoader knows which dataframe(s) to save.

These functionalities have all been added with relevant unit tests.

### DIFF
--- a/src/atc/etl/extended_loader.py
+++ b/src/atc/etl/extended_loader.py
@@ -1,0 +1,46 @@
+from typing import List
+
+from pyspark.sql import DataFrame
+
+from .types import EtlBase, dataset_group
+
+
+class ExtendedLoader(EtlBase):
+    """
+    An ExtendedLoader is a data sink, which is indicated by the fact that
+    the save method has no return.
+
+    In regards to the etl step, an ExtendedLoader USES the input dataset(s)
+    and does not consume or change it.
+    """
+
+    def __init__(
+        self, dataset_input_key: str = None, dataset_input_key_list: List[str] = None
+    ):
+        self.dataset_input_key = dataset_input_key
+        self.dataset_input_key_list = dataset_input_key_list
+
+    def etl(self, inputs: dataset_group) -> dataset_group:
+
+        if self.dataset_input_key:
+            self.save(inputs[self.dataset_input_key])
+        elif self.dataset_input_key_list:
+            self.save_many(
+                {
+                    datasetKey: df
+                    for datasetKey, df in inputs.items()
+                    if datasetKey in self.dataset_input_key_list
+                }
+            )
+        elif len(inputs) == 1:
+            self.save(next(iter(inputs.values())))
+        else:
+            self.save_many(inputs)
+
+        return inputs
+
+    def save(self, df: DataFrame) -> None:
+        raise NotImplementedError()
+
+    def save_many(self, datasets: dataset_group) -> None:
+        raise NotImplementedError()
--- a/src/atc/etl/extended_transformer.py
+++ b/src/atc/etl/extended_transformer.py
@@ -1,0 +1,58 @@
+from abc import abstractmethod
+from typing import List
+
+from pyspark.sql import DataFrame
+
+from atc.etl.types import EtlBase, dataset_group
+
+
+class ExtendedTransformer(EtlBase):
+    """If you only want to transform a single input dataframe,
+    implement `process`
+    if you want to transform a set of dataframes,
+    implement `process_many`
+
+    In regards to the etl step, the ExtendedTransformer does NOT CONSUME the inputs
+    and APPENDs the result of its transformation stage to the list of inputs.
+    """
+
+    def __init__(
+        self,
+        *,
+        dataset_output_key: str = None,
+        dataset_input_key: str = None,
+        dataset_input_key_list: List[str] = None,
+    ):
+        if dataset_output_key is None:
+            dataset_output_key = type(self).__name__
+        self.dataset_output_key = dataset_output_key
+        self.dataset_input_key = dataset_input_key
+        self.dataset_input_key_list = dataset_input_key_list
+
+    def etl(self, inputs: dataset_group) -> dataset_group:
+
+        if self.dataset_input_key:
+            df = self.process(inputs[self.dataset_input_key])
+        elif self.dataset_input_key_list:
+            df = self.process_many(
+                {
+                    datasetKey: df
+                    for datasetKey, df in inputs.items()
+                    if datasetKey in self.dataset_input_key_list
+                }
+            )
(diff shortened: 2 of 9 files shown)


src/atc/etl/extended_transformer.py
```diff
@@ -0,0 +1,57 @@
...
+from atc.etl.types import EtlBase, dataset_group
+
+
+class ExtendedTransformer(EtlBase):
+    """If you only want to transform a single input dataframe,
+    implement `process`
+    if you want to transform a set of dataframes,
+    implement `process_many`
+
+    In regards to the etl step, the transformer CONSUMES the inputs
```
JeppeBlixen, line 15: This doc is not correct, the Transformer is not consuming the input

tests/cluster/etl/test_extended_transformer.py
```diff
@@ -0,0 +1,90 @@
...
+        [
+            StructField("id", IntegerType(), False),
+            StructField("text", StringType(), False),
+        ]
+    )
+
+    return Spark.get().createDataFrame(data=data, schema=schema)
+
+
+if __name__ == "__main__":
```
LauJohansson, line 89: why do you need line 89-90?

tests/cluster/etl/test_loader.py
```diff
@@ -8,55 +8,78 @@
...
+    return Spark.get().createDataFrame(data=data, schema=schema)
+
+
+def create_empty_dataframe():
+    data = [()]
+    return Spark.get().createDataFrame(data=data)
+
 
 if __name__ == "__main__":
     unittest.main()
```
LauJohansson, line 85: why line 84 and 85?

tests/cluster/etl/test_transformer.py
```diff
@@ -25,6 +25,10 @@ def test_process_many(self):
         self.assertEqual({"TestTransformer"}, set(result.keys()))
         self.assertEqual(6, list(result.values())[0].count())
 
+    def test_dataset_key(self):
+        result = TestTransformer("my_key").etl({"df": self.df})
```
LauJohansson, line 29: create these test in a new testfile which only tests "ExtendedTransformer".

LauJohansson (changes requested):

LauJohansson:
Since it is a kind of "core class", I would suggest that documentation of the classes are added like https://github.com/atc-net/atc-dataplatform/tree/main/docs/etl

src/atc/etl/loader.py
```diff
@@ -12,15 +14,29 @@ class Loader(EtlBase):
     and does not consume or change it.
     """
 
-    def __init__(self):
-        super().__init__()
```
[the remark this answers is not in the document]
MartinBoge: The changes to the loader does not change its core logic. The code added allows the loader to call either save or save_many dependent on the dataset_input_key(s) provided. If no keys are provided it will resort to do what it has always done - checking the length of the input dataset and use this to evaluate wether calling save or save_many.
The downside to building an extended loader is that it will act entirely as the current loader - bar the ability to handle keys. Further, it would require loaders (such as the delta loader) to have two classes "per type of loader" (i.e., it would require a delta loader for the normal loader and a delta loader for the extended loader) - and they would be almost identicial except for the handling of dataset keys.

src/atc/utils/MockExtractor.py
```diff
@@ -1,15 +1,13 @@
...
 from pyspark.sql import DataFrame
 
 from atc.etl import Extractor
 
 
-class MockExtractor(Extractor, MagicMock):
-    def __init__(self, *args, dataset_key: str = None, df: DataFrame = None, **kwargs):
-        Extractor.__init__(self, dataset_key=dataset_key)
-        MagicMock.__init__(self, *args, **kwargs)
+class MockExtractor(Extractor):
```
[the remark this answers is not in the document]
MartinBoge: I will refactor this class to be "TestExtractor" instead.

tests/cluster/etl/test_extended_transformer.py
MartinBoge: I ran these locally when writing the tests to check along the way. I just left it for ease in the future - should remove?

tests/cluster/etl/test_loader.py
MartinBoge: I ran these locally when writing the tests to check along the way. I just left it for ease in the future - should remove?

tests/cluster/etl/test_extended_transformer.py
LauJohansson: In my opinion I would remove it.

tests/cluster/etl/test_loader.py
LauJohansson: In my opinion I would remove it.

tests/cluster/etl/test_transformer.py
MartinBoge: This test is just an addition to testing the functionality of the core transformer class. I noticed that there really was no test of the dataset_key functionality so i added it with this PR. Want me to remove?

LauJohansson: No - please let it stay. I thought it had something to do with the extended version. Good job!

src/atc/etl/extended_loader.py
```diff
@@ -0,0 +1,46 @@
+from typing import List
+
+from pyspark.sql import DataFrame
+
+from .types import EtlBase, dataset_group
+
+
+class ExtendedLoader(EtlBase):
+    """
+    A loader is a data sink, which is indicated by the fact that
```
LauJohansson, line 10: An extended loader*

src/atc/etl/extended_loader.py
```diff
@@ -0,0 +1,46 @@
...
+    the save method has no return.
+
+    In regards to the etl step, a loader USES the input dataset(s)
```
LauJohansson, line 13: (...), an extended loader* USES (...)

LauJohansson (approved):

LauJohansson (approved):

PR #152, opened 2022-09-12
test if deployment even works

LauJohansson:


LauJohansson:
This PR was used for testing if the deployment pipeline worked. We did a cleanup of an keyvault secret which was refering to a deleted SPN. The secret is now updated to refer to the new SPN.

PR #155, opened 2022-09-12
Feature/sql upsert improvements

JeppeBlixen:
Various improvements to sql upsert

- Refactor to use sql temp tables
- Refactor to use contextmanager, to use 'with connect as conn' syntax
- Add join col filter
- Add overwrite if target is empty
- Add documetnaion of upsert

### DIFF
--- a/tests/cluster/sql/test_deliverysql.py
+++ b/tests/cluster/sql/test_deliverysql.py
@@ -171,7 +171,20 @@ class DeliverySqlServerTests(DataframeTestCase):
         table_exists = self.sql_server.load_sql(sql_argument)
         self.assertEqual(table_exists.count(), 0)
 
-    def test16_upsert_to_table_dummy(self):
+    def test16_upsert_to_table_none_input(self):
+
+        val_return = self.sql_server.upsert_to_table_by_name(
+            df_source=None,
+            table_name=self.table_upsert_name,
+            join_cols=["testid"],
+            filter_join_cols=False,
+        )
+
+        self.assertEqual(val_return, None)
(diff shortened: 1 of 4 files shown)


tests/cluster/sql/test_deliverysql.py
```diff
@@ -213,13 +226,88 @@ def test16_upsert_to_table_dummy(self):
...
+            schema=upsertTableSchema,
+            columns=["testid", "testdata"],
+            data=[(2, "newtestdata2"), (3, "testdata3")],
+        )
+        self.sql_server.upsert_to_table_by_name(
+            df_source=df_upsertTest,
+            table_name=self.table_upsert_name,
+            join_cols=["testid"],
+            filter_join_cols=False,
+            overwrite_if_target_is_empty=False,
```
LauJohansson, line 300: The name is "test19_upsert_to_table_overwrite_empty_target" but the overwrite_if_target_is_empty=false. It this intended?

JeppeBlixen: Fixed. But it is actually very hard to fix the overwrite scenario. The end state is the same as a merge, and the temp table is cleaned.

LauJohansson:
All in all, I think it is great improvements. It alligns a lot with the delta upsert. I did not have any specific change requirements.

tests/cluster/sql/test_deliverysql.py
LauJohansson: I agree.

LauJohansson (approved):

PR #156, opened 2022-09-13
cosmos handle ready for testing

### LINKED ISSUE #134: Cosmos handle
Create a handle class for cosmos. It should inherit ThHandle, and therefor conform to the methods of DbHandle and SqlHandle.

mrmasterplan:
closes #134


tests/cluster/cosmos/test_comos.py
```diff
@@ -20,7 +20,14 @@ class CosmosTests(unittest.TestCase):
     def test_01_tables(self):
```
LauJohansson, line 20: the file name has a misspelling "tests/cluster/cosmos/test_comos.py" -> "tests/cluster/cosmos/test_cosmos.py"

src/atc/cosmos/cosmos_handle.py
```diff
@@ -0,0 +1,75 @@
...
+        rows_per_partition: int = None,
+        schema: StructType = None,
+    ):
+        self._name = name
+        self._cosmos_db = cosmos_db
+        self._rows_per_partition = rows_per_partition
+        self._schema = schema
+
+    def read(self) -> DataFrame:
+        """Read table by path if location is given, otherwise from name."""
```
LauJohansson, line 37: can it read a table by path?

LauJohansson (changes requested):

tests/cluster/cosmos/test_comos.py
```diff
@@ -20,7 +20,14 @@ class CosmosTests(unittest.TestCase):
...
         tc = TableConfigurator()
         tc.clear_all_configurations()
-        tc.register("CmsTbl", {"name": "CosmosTable"})
+        tc.register(
+            "CmsTbl",
+            {
+                "name": "CosmosTable",
+                "schema": "id string, pk string, value int",
+                "rows_per_partition": 5,
```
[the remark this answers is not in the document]
mrmasterplan: actually, no, i had to fix the table configurator.

src/atc/cosmos/cosmos_handle.py
mrmasterplan: no. bad text

src/atc/cosmos/cosmos.py
```diff
@@ -111,3 +118,51 @@ def delete_item(
...
+        db.delete_container(table_name)
+
+    def recreate_container_by_name(self, table_name: str):
+        """
+        Delete and recreate the container while preserving properties as
+        far as possible.
+        """
+        db = self.client.get_database_client(self.database)
+
+        for container in db.list_containers():
```
LauJohansson, line 136: Isnt there an easier way to ask the Cosmos db about a table than looping trhough all tables?

src/atc/cosmos/cosmos_base_server.py
```diff
@@ -0,0 +1,41 @@
...
+    ):
+        pass
+
+    def delete_container(self, table_id: str):
+        pass
+
+    def delete_container_by_name(self, table_name: str):
+        pass
+
+    def recreate_container_by_name(self, table_name: str):
```
LauJohansson, line 40: Lets be better to make docstrings for methods. For every method there is not in the BaseServer - you could consider making a small docstring.

LauJohansson: Well.. unless the method is self explaining?

src/atc/cosmos/cosmos_handle.py
```diff
@@ -0,0 +1,70 @@
...
+        self._cosmos_db.delete_container_by_name(self._name)
+
+    def drop_and_delete(self) -> None:
+        self.drop()
+
+    def write_or_append(self, df: DataFrame, mode: str) -> None:
+        if mode == "append":
+            return self.append(df)
+        else:
+            return self.overwrite(df)
```
LauJohansson, line 64: When I read this again, maybe there should be an if, elif and else.
Since it is perhaps too implicit that it overwrites per default. 

Maybe we should change the tablehandle to write_or_append to:
```
def write_or_append(self, df: DataFrame, mode: str = "overwrite") -> None:
        if mode == "append":
            return self.append(df)
       elif mode=="overwrite":
        else:
            raise ValueError("Unknown write mode")
            ```

File: src/atc/cosmos/cosmos.py [hunk 7]
```diff
@@ -12,14 +12,21 @@
...
 
 
-class CosmosDb:
+class AtcCosmosException(AtcException):
+    pass
+
+
+class CosmosDb(CosmosBaseServer):
     def __init__(
         self,
```
LauJohansson, line 31: Why dont use have a ```self.db = self.client.get_database_client(self.database)``` in your ```__init__```? It seems like you refer to it in several methods the exact same way.

src/atc/cosmos/cosmos_base_server.py
```diff
@@ -0,0 +1,41 @@
+from typing import Union
+
+from pyspark.sql import DataFrame
+from pyspark.sql.types import DataType
+
+from atc.sql import BaseServer
+
+
+class CosmosBaseServer(BaseServer):
```
LauJohansson, line 9: Shouldnt CosmosBaseServer inherit all methods from baseserver?

LauJohansson (changes requested):

LauJohansson:
@mrmasterplan - the fail message in the pipeline is weird. The graph list spn cannot find the AtcDbSpn. So the pipeline tries to create the SPN. But, since it is actually there, it fails! So could there be an error in the Graph API? or is it our deployment?

LauJohansson:
> @mrmasterplan - the fail message in the pipeline is weird. The graph list spn cannot find the AtcDbSpn. So the pipeline tries to create the SPN. But, since it is actually there, it fails! So could there be an error in the Graph API? or is it our deployment?
> 
> Using (Graph-Rest -url "servicePrincipals").value) I cannot identify the AtcDbSpn. But in Azure Portal I can see it.

We deleted the App registration, removed the secret from keyvault + purged. Then it worked again.

src/atc/cosmos/cosmos.py
mrmasterplan: well spotted. I added it.

LauJohansson:
@mrmasterplan  - I think some cleanup of testdata in the CosmosDb is needed. It also interrupts test in another PR not related to Cosmos code changes.

LauJohansson (approved):

PR #158, opened 2022-09-14
Reporting class to automatically throw exceptions to Insights

### LINKED ISSUE #157: Appinsight reporting
Set up a class that takes as its one parameter the instrumentation key of an azure app insight.

https://docs.microsoft.com/en-us/azure/azure-monitor/app/opencensus-python

You can use the snippets from here to set up the app-insights and get that key:

https://github.com/atc-net/atc-snippets/tree/main/azure-cli/monitor

The new class, maybe named `AtcInsightsReporter`, should take any arguments that are serializable, to json, and add them as properties to a log entry that is sent to the appinsights. Importantly, the class should use the job introspection available in atc-dataplatform to also add a link to the correct workspace and job-run, so that one can find the way back to the line that caused the log.

The class should be usable as a context manager, where it catches all exceptions and reports them to log analytics.
This will allow for developer attention, while also letting the job run to successful completion.

mrmasterplan:
closes #157 

Initial implementation needs testing.

### DIFF
--- a/.github/deploy-bicep/main.bicep
+++ b/.github/deploy-bicep/main.bicep
@@ -21,6 +21,8 @@ param allowUserIp string
 param sqlServerAdminUser string
 @secure()
 param sqlServerAdminPassword string
+param loganalyticsname string
+param appinsightname string
 
 // Creating permanent resource group
 module rgModule 'rg-permanent.bicep' = {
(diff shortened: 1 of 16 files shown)


mrmasterplan:
I will abandon this for now since the slack integration is simpler see #204

PR #159, opened 2022-09-16
Fix: test connection before writing to table

LauJohansson:
This PR ensures that there is a connection before writing to the SQL server.
Before: If the write_table method was called, and the server is sleeping, there is a failure
now: the connection is tested before writing to the SQL server to wake it up.


JeppeBlixen (approved):

LauJohansson:
I think some new Cosmos testing has resulting in legacy data stored in the database. This results in wrong tests.

Issue #148, opened 2022-09-07
Flaky release workflow


LauJohansson:
See this run: https://github.com/atc-net/atc-dataplatform/actions/runs/3005506556

HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/        
         File already exists. See https://pypi.org/help/#file-name-reuse for    
         more information.

mrmasterplan:
closed in #166

PR #166, opened 2022-09-27
check both test and prod for version

mrmasterplan:
for some reason the link here
[https://test.pypi.org/pypi/atc-dataplatform/json](https://test.pypi.org/pypi/atc-dataplatform/json)
currently claims the current version is 1.0.46
even though here 
[https://test.pypi.org/project/atc-dataplatform/](https://test.pypi.org/project/atc-dataplatform/)
it is 1.0.47
The post-merge pipline fails because 1.0.47 exists.
When I check the production version, at 
[https://pypi.org/pypi/atc-dataplatform/json](https://pypi.org/pypi/atc-dataplatform/json)
it says 1.0.47, so by considering the highest of either api, we will be goo again.

Another nice effect of this is that we can always re-run the release pipe to re-publish with the next higher version number now.

### DIFF
--- a/utilities/manipulate_version.py
+++ b/utilities/manipulate_version.py
@@ -13,7 +13,7 @@ version_file_path = "src/VERSION.txt"
 
 def main():
     # find out what version to use
-    pypi_v = get_test_pypi_version()
+    pypi_v = max(get_version("test.pypi.org"), get_version("pypi.org"))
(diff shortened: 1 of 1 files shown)


LauJohansson (approved):

mrmasterplan:
closes #148

PR #167, opened 2022-09-27
TimeZone Transformer

frederikgjensen:
A transformer that uses longitude and latitude values to extract timezones.

### DIFF
--- a/src/atc/transformers/timezone_transformer.py
+++ b/src/atc/transformers/timezone_transformer.py
@@ -2,6 +2,7 @@ import pyspark.sql.functions as F
 from pyspark.sql import DataFrame
 from timezonefinder import TimezoneFinder
 
+from atc.atc_exceptions import ColumnDoesNotExistException
 from atc.etl import ExtendedTransformer
 from atc.etl.types import dataset_group
 
@@ -52,10 +53,14 @@ class TimeZoneTransformer(ExtendedTransformer):
         columns = df.columns
 
         if self.latitude_col not in columns:
-            raise Exception("The specified latitude column is not in the DataFrame")
+            raise ColumnDoesNotExistException(
+                "The specified latitude column is not in the DataFrame"
+            )
 
         if self.longitude_col not in columns:
-            raise Exception("The specified longitude column is not in the DataFrame")
+            raise ColumnDoesNotExistException(
+                "The specified longitude column is not in the DataFrame"
+            )
 
         timezone_extractor = F.udf(
             lambda latitude, longitude: None
--- a/src/atc/transformers/timezone_transformer.py
+++ b/src/atc/transformers/timezone_transformer.py
@@ -2,7 +2,8 @@ import pyspark.sql.functions as F
 from pyspark.sql import DataFrame
 from timezonefinder import TimezoneFinder
 
-from atc.etl.extended_transformer import ExtendedTransformer
+from atc.etl import ExtendedTransformer
+from atc.etl.types import dataset_group
 
 
 class TimeZoneTransformer(ExtendedTransformer):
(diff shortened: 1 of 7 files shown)


src/atc/transformers/timezone_transformer.py
```diff
@@ -0,0 +1,62 @@
...
+    process(df: DataFrame):
+        returns the input DataFrame with a TimeZone column
+    """
+
+    def __init__(
+        self,
+        longitude_col: float,
+        latitude_col: float,
+        dataset_input_key: str,
+        dataset_output_key: str,
```
LauJohansson, line 36: dataset_input_key and data_output_key are optional in the extended transformer. Is it on purpurse that they are required here?

src/atc/transformers/timezone_transformer.py
```diff
@@ -0,0 +1,62 @@
+import pyspark.sql.functions as F
+from pyspark.sql import DataFrame
+from timezonefinder import TimezoneFinder
+
+from atc.etl.extended_transformer import ExtendedTransformer
```
LauJohansson, line 5: Thinks that it is possible to write: "from atc.etl import ExtendedTransformer"

tests/local/transformers/test_timezone_transformer.py
```diff
@@ -0,0 +1,37 @@
...
+class TimeZoneTransformerTest(DataframeTestCase):
+    def test_timezone_transformer(self):
+        input_schema = T.StructType(
+            [
+                T.StructField("Latitude", T.DoubleType(), True),
+                T.StructField("Longitude", T.DoubleType(), True),
+            ]
+        )
+
+        input_data = [(51.519487, -0.083069), (55.6761, 12.5683)]
```
LauJohansson, line 17: also test the None logic. Where either lat or lon are None.

tests/local/transformers/test_timezone_transformer.py
```diff
@@ -0,0 +1,37 @@
...
+        expected_data = [
+            (51.519487, -0.083069, "Europe/London"),
+            (55.6761, 12.5683, "Europe/Copenhagen"),
+        ]
+
+        self.assertDataframeMatches(
+            df=transformed_df,
+            columns=None,
+            expected_data=expected_data,
+        )
```
LauJohansson, line 37: What happens if some of the lat/lon values exeeds the maximum/minium expected values?

src/atc/transformers/timezone_transformer.py
```diff
@@ -0,0 +1,62 @@
...
+        )
+        self.longitude_col = longitude_col
+        self.latitude_col = latitude_col
+        self.column_output_name = column_output_name
+
+    def process(self, df: DataFrame) -> DataFrame:
+        """
+        Extracts a timezone using latitude and longitude columns,
+        and outputs the input DataFrame with a TimeZone column.
+        """
```
LauJohansson, line 50: Maybe throw exeption if the longitude_col or latitude_col is not in the dataframe?

LauJohansson (changes requested):

tests/local/transformers/test_timezone_transformer.py
frederikgjensen: ValueError: The given latitude 120.2 is out of bounds

src/atc/transformers/timezone_transformer.py
```diff
@@ -0,0 +1,74 @@
...
+        columns = df.columns
+
+        if self.latitude_col not in columns:
+            raise Exception("The specified latitude column is not in the DataFrame")
```
LauJohansson, line 55: To further improve your code you could consider adding something like this to the "atc_exceptions.py" file. 

```python
class ColumnDoesNotExistException(AtcException):
    pass
```
Then import it to the file:
```python
from atc.atc_exceptions import ColumnDoesNotExistException
```

And then use it as 
```python
raise ColumnDoesNotExistException("The specified latitude column is not in the DataFrame")
```

This way it becomes easier for the uers to understand the exeptions. We can then reuse the exeption in other classes too.

LauJohansson (changes requested):

LauJohansson (approved):

PR #168, opened 2022-09-28
feat: use tablehandle in upsertloader

LauJohansson:
The UpsertLoader now uses tablehandle so it works with e.g. both delta and sql server.

The logic is almost identical to the upsert logic in DeltaHandle().upsert.
If one want to do a full load, the target table should be empty.


tests/cluster/etl/test_upsertloader.py
```diff
@@ -5,14 +5,12 @@
 from atc.config_master import TableConfigurator
 from atc.delta import DbHandle, DeltaHandle
 from atc.etl.loaders.UpsertLoader import UpsertLoader
-from atc.etl.loaders.UpsertLoaderParameters import UpsertLoaderParameters
 from atc.utils import DataframeCreator
 from tests.cluster.delta import extras
 from tests.cluster.delta.SparkExecutor import SparkSqlExecutor
 
 
 class UpsertLoaderTests(DataframeTestCase):
```
JeppeBlixen, line 13: I don't see the need for test_02_can_perform_full_over_existing.
It should be possible to do and incremental and merge test from the state from test_01_can_perform_incremental_on_empty.

This would also make the data objects data1 to data4 a bit simpler, and therefore also have a simpler data5 object after the merge

src/atc/etl/loaders/UpsertLoader.py
```diff
@@ -1,101 +1,23 @@
...
-            insert_cols=df.columns,
-            update_cols=non_join_cols,
-            special_update_set="",
-        )
-
-        Spark.get().sql(merge_sql_statement)
-
-        print("Incremental Base - incremental load with merge")
+        """Upserts each dataset in the datasetgroup individually."""
+        for df in datasets.values():
```
mrmasterplan, line 18: To me this is surprising behavior. If I want this behavior, I would inject a UnionTransformer before the loader. I suggest we remove the save_many functionality completely.

mrmasterplan (changes requested):

src/atc/etl/loaders/UpsertLoader.py
```diff
@@ -1,101 +1,21 @@
...
-            target_table_name=target_table_name,
-            source_table_name=temp_view_name,
-            join_cols=self.params.join_cols,
-            insert_cols=df.columns,
-            update_cols=non_join_cols,
-            special_update_set="",
-        )
-
-        Spark.get().sql(merge_sql_statement)
+        raise NotImplementedError()
```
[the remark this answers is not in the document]
LauJohansson: I have created a loop which upserts all the provided dfs to a target table

tests/cluster/etl/test_upsertloader.py
LauJohansson: I have removed test 02. And cleaned up the test. Then I have added a test of the save many.

src/atc/etl/loaders/UpsertLoader.py
JeppeBlixen: I get your point Simon.

We only use the save_many in a single loader called OrchestratorLoader. The OrchestratorLoader is used to pass an output of one orchestrator as input to a second orchestrator. Here we have the save_many that pass the dataset into the execute function. It could we argued that this Loader could be its own class based step. Then i don't see a use for the save_many on a Loader.

LauJohansson: I will remove the save many method. Also since it is much less efficient to upsert each dataframe seperately instead of using the UnionTransformer first - then call save().

JeppeBlixen (approved):

mrmasterplan (approved):

PR #176, opened 2022-10-07
Include dataset keys from ExtendedLoader into Loader

JeppeBlixen:
Removing ExtendedLoader and moving the features regarding dataset keys into the Loader, as this is backwards comparable.
Updates docs and tests.


tests/local/utils/test_mock_etl.py
```diff
@@ -37,3 +37,7 @@ def process(self, df: DataFrame) -> DataFrame:
     def test_mocking(self):
         load = MockLoader()
         load.hello.world.foo.bar()
+
+
+if __name__ == "__main__":
```
LauJohansson, line 42: I think this is unneccesary

src/atc/etl/loader.py
```diff
@@ -12,15 +14,23 @@ class Loader(EtlBase):
     and does not consume or change it.
     """
 
-    def __init__(self):
-        super().__init__()
+    def __init__(
+        self, dataset_input_key: str = None, dataset_input_key_list: List[str] = None
+    ):
```
LauJohansson, line 19: What about super()__init(). ?

tests/cluster/etl/test_loader.py
```diff
@@ -9,41 +9,100 @@
 
 class LoaderTests(unittest.TestCase):
     @classmethod
-    def setUpClass(cls):
+    def setUp(self):
```
LauJohansson, line 12: why not use setUpClass?

tests/cluster/etl/test_loader.py
```diff
@@ -9,41 +9,100 @@
...
-        # assert that loader returns something:
-        self.assertIs(input, result)
+        self.assertFalse(self.loader.save.called)
+        self.assertTrue(self.loader.save_many.called)
 
-        args = self.loader.save.call_args[
+        args = self.loader.save_many.call_args[
             0
         ]  # the .args part became available in python 3.8
         print(args)
```
LauJohansson, line 66: is it neccesary to print the args?

tests/cluster/etl/test_loader.py
```diff
@@ -9,41 +9,100 @@
...
+            {"my_key_1": self.df_1, "my_key_2": self.df_2, "my_key_3": self.df_3}
+        )
+
+        self.assertFalse(self.loader_input_key_list.save.called)
+        self.assertTrue(self.loader_input_key_list.save_many.called)
+
+        args = self.loader_input_key_list.save_many.call_args[
             0
         ]  # the .args part became available in python 3.8
         print(args)
```
LauJohansson, line 101: is it neccesary to print the args?

tests/local/utils/test_mock_etl.py
LauJohansson: if it inherits pytest and has the prefix *test_* it will be executed. See https://docs.pytest.org/en/7.1.x/how-to/usage.html

LauJohansson (changes requested):

tests/local/utils/test_mock_etl.py
JeppeBlixen: It is not for the test to run. It is so i can run the script in the debugger.

tests/cluster/etl/test_loader.py
JeppeBlixen: It is because i validate if the functions have been called:
Save test:
self.assertTrue(self.loader.save.called)
self.assertFalse(self.loader.save_many.called)
Save_many test:
self.assertFalse(self.loader.save.called)
self.assertTrue(self.loader.save_many.called)

This would not work for setupClass, as the function would always be called

tests/cluster/etl/test_loader.py
JeppeBlixen: No, it was just already there, i will remove.

tests/cluster/etl/test_loader.py
JeppeBlixen: No, it was just already there, i will remove.

src/atc/etl/loader.py
JeppeBlixen: Fixed

tests/cluster/etl/test_loader.py
LauJohansson: Alright. At first look I imagined that the classes (and datasets) what setup too many times. But I see now.

tests/local/utils/test_mock_etl.py
LauJohansson: Since it is for local running in the debugger -maybe it should beremoved before merging? 
I have made the same comment before: https://github.com/atc-net/atc-dataplatform/pull/149

LauJohansson: If it is a good thing to use - then we should consider use this in all tests-scripts? 
I think it is just my lack of understanding of what it does.

JeppeBlixen: It is fair, i will remove it.

LauJohansson (commented):
I think it look fine. Do we need two more eyes on this, since it makes changes (though backward compatible) to a core class?

src/atc/etl/loader.py
```diff
@@ -12,15 +14,29 @@ class Loader(EtlBase):
     and does not consume or change it.
     """
 
-    def __init__(self):
+    def __init__(
+        self, dataset_input_key: str = None, dataset_input_key_list: List[str] = None
```
mrmasterplan, line 18: I dislike that it is possible to set both `dataset_input_key` and `dataset_input_key_list`. The behavior for a list of length 1 is the same as for a single key.

I recomment that the function should have the following signature 
```
def __init__(self, *dataset_input_keys: str):
```
this gives a list of string with length 0, 1 or more. 0 keys is consuming, 1 key results in a `save` and more keys is a `save_many`. This would make it unique, but still flexible.

mrmasterplan: if you don't like the variadic style, then I propose to still only use the `dataset_input_key_list` which can optionally have a single key.

mrmasterplan (changes requested):

src/atc/etl/loader.py
JeppeBlixen: I get you'r point. The reason why there are two parameters is to have full control of the save vs. save_many. When setting dataset_input_key i know save is called and when dataset_input_key_list i know save_many is called.
This is maybe not the best reason.

What about:
def __init__(self, dataset_input_keys: Union[str, List[str]] = None):

dataset_input_keys can both be string and list of strings, and save/save_many is called accordingly.
If dataset_input_keys is None, the original behaviour is used.

LauJohansson: @JeppeBlixen  - this PR has been here for some time. Have you implemented the above mentioned changes? ordo you need any assistance?

JeppeBlixen: Done

mrmasterplan (approved):

LauJohansson (approved):

PR #179, opened 2022-10-17
Non-consuming transformer refactor

MartinBoge:
Scope of this PR: To align the parameterization of the non-consuming transformer base class with the base class of the loader. 

Moreover, having the TransformerNC base class take the parameter "dataset_input_keys" instead of both "dataset_input_key" and "dataset_input_key_list". 
The new parameter "dataset_input_keys" is a union, such that it takes either a string or a list of strings. 

This change is not backward compatible, thus it calls for a an upgrade of the minor version from 1.0.* to 1.1.*


mrmasterplan (approved):
Looks like a good improvement overall.

PR #185, opened 2022-10-19
delta handle can now own schema and partitioning

mrmasterplan:


LauJohansson:
I can see the pipeline still has issues. I think that we might need to wait a day to see if the API is fixed. Maybe it does not has something to do with the databricks API but with the azure resource. The error looks like it related to azure storage. Maybe the location of the odbc driver cannot be found if theres is an issue with the storage resource.

LauJohansson:
Here is a guess of what could be wrong: if the there is some issue with the azure resource manager or at least the databricks workspace resource - then, when trying to invoke the global-init-script API it is missing something?

LauJohansson:
Might coordinate with #188 ?

mrmasterplan:
this will be andled in a number of other PR. substantial rewrite will be handled in new PR

PR #186, opened 2022-10-20
feat: Add odbc driver using python

LauJohansson:
This PR replaces the use of databricks API to download the ODBC driver. Instead use python in setup job.


mrmasterplan:
is this still relevant?

PR #187, opened 2022-10-20
use new init script setup

mrmasterplan:
Do it as in example https://learn.microsoft.com/en-us/azure/databricks/clusters/init-scripts#example-install-postgresql-jdbc-driver


mrmasterplan:
method does not work. abandoning for now

PR #191, opened 2022-10-27
partitioning from delta handle

mrmasterplan:
Get partitioning from delta handle in a nice format.


src/atc/delta/delta_handle.py
```diff
@@ -110,6 +112,35 @@ def recreate_hive_table(self):
...
+        +-----------------+---------------+-------+
+        |         col_name|      data_type|comment|
+        +-----------------+---------------+-------+
+        |           mycolA|         string|       |
+        |           myColB|            int|       |
+        |                 |               |       |
+        |   # Partitioning|               |       |
+        |           Part 0|         mycolA|       |
+        +-----------------+---------------+-------+
+        but thie method return the partitioning in the form ['mycolA']
```
LauJohansson, line 126: misspelling

src/atc/delta/delta_handle.py
```diff
@@ -110,6 +112,35 @@ def recreate_hive_table(self):
...
+        """
+        if self._partitioning is None:
+            rows_iter = iter(
+                Spark.get().sql(f"DESCRIBE TABLE {self.get_tablename()}").collect()
+            )
+            for row in rows_iter:
```
LauJohansson, line 132: could you provide a # comment on what this logic does?

src/atc/delta/delta_handle.py
```diff
@@ -110,6 +112,35 @@ def recreate_hive_table(self):
         self.drop()
         self.create_hive_table()
 
+    def get_partitioning(self):
```
LauJohansson, line 115: what if there is no partion? Is that even possible?

LauJohansson: if it is - how does the method handle it?

LauJohansson (dismissed):

src/atc/delta/delta_handle.py
mrmasterplan: its possible, returns empty list. I added a test.

src/atc/delta/delta_handle.py
mrmasterplan: added comments.

src/atc/delta/delta_handle.py
mrmasterplan: fixed

LauJohansson (approved):

Okami1 (commented):
Ser godt ud, ingen kommentarer for ændringer

PR #194, opened 2022-10-31
preview json

mrmasterplan:


LauJohansson (approved):

src/atc/orchestrators/ehjson2delta/EhJsonToDeltaOrchestrator.py
```diff
@@ -8,6 +10,7 @@
 )
 
 
+@deprecation.deprecated(details="Currently in preview.")
```
LauJohansson, line 13: is there any other which are non-tested? those should might also have preview in the deprecation details.

LauJohansson (changes requested):

LauJohansson (dismissed):

LauJohansson (approved):

mrmasterplan:
I have tested the full Orchestrator on a real load. It works as expected. Therefore there is no longer any need to mark it as preview. I am rejecting this PR.

PR #197, opened 2022-11-01
feat: no sqlexecutor split on semicolon

LauJohansson:


mrmasterplan:
https://stackoverflow.com/questions/36938399/how-to-pass-multiple-statements-into-spark-sql-hivecontext

https://thomasdecaux.medium.com/multi-statement-with-spark-sql-fc9996724576

http://mark.thegrovers.ca/tech-blog/spark-sql-gotchas

this will probably not work.

PR #202, opened 2022-11-10
Skip the wait in Sql Server deployment

mrmasterplan:
Step 1 introduce the error and see at which step it fails.


.github/deploy/steps/65-Provision-Spn-db-user.ps1
```diff
@@ -10,7 +10,8 @@ Graph-CreateRole -principalId $spId -roleDefinitionId 88d8e3e3-8f55-4a1e-953a-9b
 Ignore-Errors  # The role may already be there if we re-deploy
 
 Write-Host "Waiting for role to settle...." -ForegroundColor DarkYellow
-Start-Sleep -seconds 60 # If the Directory Reader is not there, increase te seconds here.
+#Start-Sleep -seconds 60 # If the Directory Reader is not there, increase te seconds here.
+#skip this and let's see where it fails
```
LauJohansson, line 14: Azure AD is a bit slow sometimes. The SQL server identity needed "Directory Reader" role. That takes apparently some time before AD gives the role.

PR #207, opened 2022-11-23
fix: rename sparkexecutor test file

LauJohansson:
The spark executor test had a misspelling in the file name.


LauJohansson:
@mrmasterplan - this test actually fails at: *File "/dbfs/test/.../tests/cluster/delta/test_sparkexecutor.py", line 42, in test_can_execute*. 
I can see that you have made this test (that fails): 

`self.assertEqual(Spark.get().sql('SHOW DATABASES LIKE "my_db2*"').count(), 0)`
`AssertionError: 1 != 0`

PR #210, opened 2022-11-25
Feature/configurator cli

mrmasterplan:
## The Configurator Command Line Interface (CLI)

The table configurations, available to the configurator can be useful when executing 
actions on the command line. See below for individual commands. To expose the 
configurator command line interface, you need to call the `.cli()` method *after* 
you have initialized the configurator with your project details. You therefor need 
to expose a command line script as shown below.

In the file that initializes your configurator:
```
def init_my_configurator():
    c= Configurator()
    c.add_resource_path(my_yaml_module)
    c.register('ENV', config.my_env_name)
    return c

if __name__ == "__main__":
    init_my_configurator().cli()
```
Now, all the functionality below will become available on the command line.

### Generated Keys File
When using an IDE to develop python code, a useful feature is auto-completion and 
linting. Such features are however not available when using string keys from yaml 
files. It can therefore be useful to extract the keys from the yaml configurations 
and make them available as python objects.

Call the following command to generate such a keys file from your initialized 
configurator.

```
$> my_config generate-keys-file -o keys.py
```

This will create the file `keys.py` with the following example contents:
```python
# AUTO GENERATED FILE.
# contains all atc.Configurator keys

MyFirst = "MyFirst"
MySecond = "MySecond"
MyAlias = "MyAlias"
MyForked = "MyForked"
MyRecursing = "MyRecursing"
```

You can now use the keys file to auto-complete and validate your yaml keys:
```python
DeltaHandle.from_tc("MyFirst")
```
becomes
```python
from keys import MyFirst
DeltaHandle.from_tc(MyFirst)
```
which also supports flake8 linting for correct spelling.

If you want to check that you did not forget to update the keys file as part of your 
CICD pipeline, running the same command with the additional option `--check-only` 
will return an exit code of 0 if the file was already up-to-date and 1 otherwise.


src/atc/configurator/_cli/generate_keys_file.py
```diff
@@ -0,0 +1,62 @@
+import argparse
+import os.path
+from io import StringIO
+
+from atc.exceptions.cli_exceptions import AtcCliCheckFailed
+
+
+def setup_parser(parser: argparse.ArgumentParser):
+    parser.set_defaults(func=generate_keys_file)
+    parser.add_argument("-o", "--output_file", type=str, default="")
```
LauJohansson, line 10: output-file

src/atc/configurator/_cli/generate_keys_file.py
```diff
@@ -0,0 +1,62 @@
...
+        "-c",
+        "--check-only",
+        dest="check",
+        action="store_true",
+        help="Only check, don't update the output file",
+    )
+    parser.set_defaults(check=False)
+
+
+def generate_keys_file(self, options):
```
LauJohansson, line 21: docstring comment on what this def does

tests/local/configurator/test_configurator_cli.py
```diff
@@ -0,0 +1,64 @@
...
+            c.cli()
+            # check passes since the file contents were ok
+
+            with open(name, "w") as f:
+                f.write(expected[:-10])  # bad contents
+
+            sys.argv = ["mycliprog", "generate-keys-file", "-c", "-o", name]
+            with self.assertRaises(AtcCliCheckFailed):
+                # file had bad contents
+                c.cli()
```
LauJohansson, line 64: Have you planned to do more tests?

src/atc/configurator/_cli/generate_keys_file.py
```diff
@@ -0,0 +1,62 @@
...
+    try:
+        import black
+
+        new_conts = black.format_file_contents(
+            new_conts, fast=False, mode=black.FileMode()
+        )
+    except ModuleNotFoundError:
+        pass
+
+    if not options.output_file:
```
LauJohansson, line 41: # output to screen if no output file is given

src/atc/configurator/_cli/generate_keys_file.py
```diff
@@ -0,0 +1,62 @@
...
+        print(new_conts, end="")
+        return 0
+    else:
+        if options.check:
+            if os.path.exists(options.output_file):
```
LauJohansson, line 46: fail fast pattern instead

src/atc/configurator/_cli/generate_keys_file.py
```diff
@@ -0,0 +1,62 @@
...
+import os.path
+from io import StringIO
+
+from atc.exceptions.cli_exceptions import AtcCliCheckFailed
+
+
+def setup_parser(parser: argparse.ArgumentParser):
+    parser.set_defaults(func=generate_keys_file)
+    parser.add_argument("-o", "--output_file", type=str, default="")
+    parser.add_argument(
```
LauJohansson, line 11: comment

LauJohansson (dismissed):

LauJohansson (approved):

PR #211, opened 2022-11-28
feat: Add transformer for selecting and casting columns based on schema

JeppeBlixen:
Transformer for selecting and casting columns based on pyspark schema


mrmasterplan:
the code is solid, and nice documentation.
We already have a piece of code that does something very similar. 
https://github.com/atc-net/atc-dataplatform/blob/main/src/atc/transformers/simple_sql_transformer.py
This seems to be a generalization of that.
please either 
- make a solution that works for both and change SimpleSqlServerTransformer to make use of your generalized solution, 
- or just create an issue to mark the above for future work, in that case we can just merge yours as is.

src/atc/transformers/select_and_cast_columns_transformer_nc.py
```diff
@@ -0,0 +1,41 @@
+from typing import List, Union
+
+import pyspark.sql.functions as F
+import pyspark.sql.types as T
+from pyspark.sql import DataFrame
+
+from atc.etl import TransformerNC
+
+
+class SelectAndCastColumnsTransformerNC(TransformerNC):
```
mrmasterplan, line 10: Please consider generalizing, see my comment above

mrmasterplan (changes requested):

LauJohansson:
> the code is solid, and nice documentation. We already have a piece of code that does something very similar. https://github.com/atc-net/atc-dataplatform/blob/main/src/atc/transformers/simple_sql_transformer.py This seems to be a generalization of that. please either
> 
> * make a solution that works for both and change SimpleSqlServerTransformer to make use of your generalized solution,
> * or just create an issue to mark the above for future work, in that case we can just merge yours as is.
> 
> **Edit:** Comparing the existing code to yours, I can see that the SimpleSqlServerTransformer makes use of an alias call. Since your tests pass it may not be necessary, but maybe better just in case?
> 
> And while we are on the subject of aliassing: We have seen the following transformer to be very useful. It ensures that the capitalization is as expected in the target, while at the same time being case insensitive on the input dataframe side. This could easily be incorporated into your transformer to make it even more broadly useful.
> 
> ```
>     @staticmethod
>     def downselect_columns(df: DataFrame, desired_columns: List[str]) -> DataFrame:
>         """given the input dataframe and desired columns, select those columns
>         that appear in both inputs, using the capitalization of the second input,
>         while performing case-insensitive matching."""
>         in_columns = [col.lower() for col in df.columns]
>         desired_and_available_columns = []
>         for col in desired_columns:
>             low_col = col.lower()
>             if low_col in in_columns:
>                 desired_and_available_columns.append(col)
> 
>         # select is implicitly case-insensitive
>         return df.select(*desired_and_available_columns)
> ```

I think that we have experienced, that when upgrading to a newer spark version, the .cast() without a following .alias() gave some troubles.

src/atc/transformers/select_and_cast_columns_transformer_nc.py
```diff
@@ -0,0 +1,41 @@
...
+    ):
+        super().__init__(
+            dataset_input_keys=dataset_input_keys,
+            dataset_output_key=dataset_output_key,
+        )
+        self.schema = schema
+
+    def process(self, df: DataFrame) -> DataFrame:
+
+        selectAndCastColumnsList = [F.col(c.name).cast(c.dataType) for c in self.schema]
```
LauJohansson, line 39: what happens, if the self.schema has one or more columns which are not in the df? 
How should this be handled? or should it even be possible to give a schema where some of the columns are not in the df?

LauJohansson, line 39: Consider using alias: `selectAndCastColumnsList = [F.col(c.name).cast(c.dataType).alias(cname) for c in self.schema]`. I have sometimes experienced issues with casting without aliassing (when upgrading to newer version of spark).

JeppeBlixen: In this case i have added a None column of correct type: F.lit(None).cast(c.dataType).alias(c.name)

src/atc/transformers/select_and_cast_columns_transformer_nc.py
JeppeBlixen: I have added the ability for case-insensitive matching and aligned with simple_sql_transformer.

I case we in future add the sec trunc feature to the sql loader directly, the simple_sql_transformer could be deprecated. I will make an issue on this.

src/atc/utils/SelectAndCastColumns.py
```diff
@@ -0,0 +1,36 @@
...
+            else F.lit(None).cast(c.dataType).alias(c.name)
+            for c in schema
+        ]
+    else:
+        selectAndCastColumnsList = [
+            F.col(c.name).cast(c.dataType).alias(c.name)
+            if c.name in df.columns
+            else F.lit(None).cast(c.dataType).alias(c.name)
+            for c in schema
+        ]
```
LauJohansson, line 34: I think that there is some redundant code here. I could not figure out a good way to avoid the redundant code without making a lot of new code. But here is an outline. You can use it if you feel for it! I dont know if it is good code practice....

```python
def _lower(col, caseInsensitiveMatching):
   return col.lower() if caseInsensitiveMatching else col

selectAndCastColumnsList = [
            F.col(c.name).cast(c.dataType).alias(c.name)
            if _lower(c.name,caseInsensitiveMatching) in [_lower(col,caseInsensitiveMatching) for col in df.columns]
            else F.lit(None).cast(c.dataType).alias(c.name)
            for c in schema
        ]

```

mrmasterplan (approved):

PR #214, opened 2022-12-07
Feature/configure from sql

mrmasterplan:
### Configuration from sql

When using sql statements to create and manage tables, it can be good to keep all
information referring to a table within one file. Therefore, the configurator employs 
a sql parsing library to extract table details directly from the CREATE statements. 
It can be used like this:

```python
from . import my_sql_folder
Configurator().add_sql_resource_path(my_sql_folder)
```

A sql CREATE statement already contains all information that the configurator needs 
except the key, aka. table_id, that shall be used to refer to the table.
This key needs to be added to the sql in the form of a comment with a magic prefix
`"-- atc.Configurator "` (note the spaces. not case-sensitive.) See the following 
example:
```sparksql
-- atc.Configurator key: MySparkDb
CREATE DATABASE IF NOT EXISTS `my_db1{ID}`
COMMENT "Dummy Database 1"
LOCATION "/tmp/foo/bar/my_db1/";

-- atc.Configurator key: MyDetailsTable
CREATE TABLE IF NOT EXISTS `my_db1{ID}.details`
(
  {MyAlias_schema},
  another int
  -- comment with ;
)
USING DELTA
COMMENT "Dummy Database 1 details"
LOCATION "/{MNT}/foo/bar/my_db1/details/";

-- pure configurator magic in this statement
-- atc.Configurator key: MyAlias
-- atc.Configurator alias: MySqlTable
;


-- ATC.CONFIGURATOR key: MySqlTable
-- atc.Configurator delete_on_delta_schema_mismatch: true
CREATE TABLE IF NOT EXISTS `{MySparkDb}.tbl1`
(
  a int,
  b int,
  c string,
  d timestamp
)
USING DELTA
COMMENT "Dummy Database 1 table 1"
LOCATION "/{MNT}/foo/bar/my_db1/tbl1/";
```

The example is quite complex and demonstrates a number of features:
- the magic tag "-- ATC.CONFIGURATOR " is case-insensitive
- the key has to be specified as `-- ATC.CONFIGURATOR key: MyKey`.
- any other attribute can be specified after the same magic tag and will be 
  available through the configurator. In fact, all comments with the magic tag will 
  be collected together and will be interpreted as a `yaml` document, so more 
  complicated structures are also possible.
- all python string substitutions that the configurator offers are also available 
  here in the sql configuration.
- string substitution is available inside the schema, this allows schemas to be 
  constructed from other schemas.


src/atc/configurator/_parse_sql.py
```diff
@@ -0,0 +1,232 @@
...
+    yaml_document = "\n".join(collected_lines)
+    return yaml.load(yaml_document, Loader=yaml.FullLoader) or {}
+
+
+@dataclass
+class StatementBlocks:
+    schema: str = None
+    using: str = None
+    options: str = None
+    partitioned_by: str = None
```
mrmasterplan, line 47: completely unpack the values

src/atc/configurator/_parse_sql.py
```diff
@@ -0,0 +1,232 @@
...
+        val: str = token.value.strip()
+        if not _magic_comment_start_re.match(val):
+            continue
+        collected_lines.append(val[len(_magic_comment_start) :])
+    yaml_document = "\n".join(collected_lines)
+    return yaml.load(yaml_document, Loader=yaml.FullLoader) or {}
+
+
+@dataclass
+class StatementBlocks:
```
LauJohansson, line 43: Does it also take delta live tables into account?

src/atc/configurator/sql/__init__.py
```diff
@@ -0,0 +1,88 @@
+from .sqlparse import engine, keywords, tokens
+from .sqlparse.engine import grouping
+from .sqlparse.lexer import Lexer
+
+lex = Lexer()
+lex.clear()
+
+my_regex = [
+    # (r"ZORDER\s+BY\b", tokens.Keyword),
+    (r"IF\s+NOT\s+EXISTS\b", tokens.Keyword),
```
LauJohansson, line 10: Delta live tables?

src/atc/configurator/configurator.py
```diff
@@ -11,6 +10,8 @@
 
 from atc.exceptions import NoSuchValueException
```
LauJohansson, line 12: If the sql parser is dependent on a spark version - maybe we need to assert on it?
I guess some of the SQL syntax only applies for some versions.

src/atc/configurator/sql/__init__.py
```diff
@@ -0,0 +1,88 @@
+from .sqlparse import engine, keywords, tokens
+from .sqlparse.engine import grouping
+from .sqlparse.lexer import Lexer
+
+lex = Lexer()
+lex.clear()
+
+my_regex = [
+    # (r"ZORDER\s+BY\b", tokens.Keyword),
```
mrmasterplan, line 9: copied line

src/atc/configurator/_parse_sql.py
```diff
@@ -0,0 +1,232 @@
...
+    partitioned_by: str = None
+    clustered_by: str = None
+    sorted_by: str = None
+    location: str = None
+    comment: str = None
+    tblproperties: str = None
+    as_: str = None
+
+
+def _extract_optional_blocks(stmt) -> StatementBlocks:
```
LauJohansson, line 56: input datatype missing

src/atc/configurator/_parse_sql.py
```diff
@@ -0,0 +1,232 @@
...
+            if not into_token.value.upper() == "INTO":
+                AtcConfiguratorInvalidSqlException("expected INTO after clustering")
+
+            num_buckets = int(next(stmt).value)
+
+            buckets_token = next(stmt)
+            if not buckets_token.value.upper() == "BUCKETS":
+                AtcConfiguratorInvalidSqlException("expected BUCKETS after INTO n")
+
+            cluster_line += f"\nINTO {num_buckets} BUCKETS"
```
mrmasterplan, line 114: forgot to save the wole thing into blocks

src/atc/configurator/_parse_sql.py
```diff
@@ -0,0 +1,232 @@
...
+        elif val == ";":
+            break
+        else:
+            raise AtcConfiguratorInvalidSqlException(
+                "Unknown statement form encountered."
+            )
+    return blocks
+
+
+def _walk_create_statement(statement) -> Dict:
```
LauJohansson, line 137: input datatype missing

src/atc/configurator/sql/__init__.py
mrmasterplan: not supported yet

src/atc/configurator/_parse_sql.py
mrmasterplan: implemented

src/atc/configurator/_parse_sql.py
mrmasterplan: added

src/atc/configurator/_parse_sql.py
mrmasterplan: all blocks now implemented and unit-tested

src/atc/configurator/_parse_sql.py
mrmasterplan: all unpacked now

LauJohansson (approved):

PR #218, opened 2022-12-12
enable spark 10.4 testing

mrmasterplan:
# Note to the reviewer

This PR enables testing on databricks runtime version 10.4 LTS. The following actions were taken to enable this:
- The pipeline was extended to submit a job with runtime version 10.4
- the job submission was extended with the ability to specify the dependencies, since they differ for 10.4
- the setup job had its dependencies removed, since this was complicated wnad it didn't need them
- Since now multiple test jobs are running in parallel, we must make sure they don't interfere
  - all tables have had `{ID}` added to both name and path unless they were already there
  - the cosmos class was exteded with a unique catalog name
- the cache loaded was fixed with the logic actually differing depending on the spark version. I couldn't find a single solution that works in both 9.1 and 10.4, so I had to fork.


.github/workflows/pre-integration.yml
```diff
@@ -76,20 +76,21 @@ jobs:
...
+            -sparkVersion "10.4.x-scala2.12" `
+            -testJobDetails job104.json `
+            -sparkLibs sparklibs104.json
 
       - name: Wait 2 min for things to settle
         shell: pwsh
         run: Start-Sleep -s 120
 
-      - name: Fetch integration tests 9.1
+      - name: Fetch integration tests 10.4
```
LauJohansson, line 91: is there a need to fetch integrationtests from 9.1 too?

mrmasterplan: indeed there is

LauJohansson: I agree.

LauJohansson (approved):

JeppeBlixen (approved):

Issue #219, opened 2022-12-13
Schemamanager is not compatible with Databricks connect


LauJohansson:
The Schema manager gives this error when using databricks-connect

´´´python
sc._jvm.org.apache.spark.sql.api.python.PythonSQLUtils.parseDataType(type_str).json())
´´´

mrmasterplan:
solved in #220

PR #220, opened 2022-12-13
pure python schema parsing

mrmasterplan:
I implemented the parsing of Spark SQL schemas in pure python.
Comparing the performance of the python parser as compared to delegating the task to the spark function shows a 20 fold improvement when running in a databricks notebook


src/atc/sql/schema.py
```diff
@@ -1,16 +1,315 @@
...
+        if it.peek() == ",":
+            break  # no more metadata
+
+        if it.peek() != " ":
+            raise SchemaExtractionError("Malformed SQL")
+        next(it)  # consume space
+
+        token = _get_ascii_token(it).upper()
+        if token == "NOT":
+            for c in " NULL":
```
LauJohansson, line 81: is it on purpose that the " NULL" has a leading whitespace?

src/atc/sql/schema.py
```diff
@@ -1,16 +1,315 @@
...
+                try:
+                    if next(it).upper() != c:
+                        raise SchemaExtractionError("Malformed SQL")
+                except StopIteration:
```
LauJohansson, line 85: if we could give a hint on what is wrong with the malformatting- should we print it? It would make it easier for developers to debug. This comment also applies for the other usages of `raise SchemaExtractionError("Malformed SQL")`

LauJohansson: like the goo hint in line 170

src/atc/sql/schema.py
```diff
@@ -1,16 +1,315 @@
...
+        SchemaExtractionError("Invalid start of identifier sequence.")
+
+    non_escaped_characters = string.ascii_letters + string.digits + "_"
+    non_whitespace = non_escaped_characters + string.punctuation
+
+    while True:
+        char = it.peek()
+
+        if char == "`":
+            escaped = not escaped
```
LauJohansson, line 298: what happens here with escaped = not escaped?

src/atc/schema_manager/schema_manager.py
```diff
@@ -61,7 +62,7 @@ def get_schema(self, schema_identifier: str) -> T.StructType:
 
         # Check if the schema is a sql string
         if key == "sql":
-            parsed_schema = T._parse_datatype_string(value)
```
LauJohansson, line 64: This could fix this bug #219

src/atc/sql/schema.py
```diff
@@ -1,16 +1,315 @@
...
+import re
+import string
+
+from more_itertools import peekable
 from pyspark.sql import types as t
 
-from atc.spark import Spark
 
+class SchemaExtractionError(Exception):
+    pass
```
LauJohansson, line 10: Do the parser handle delta live tables?

mrmasterplan: no

src/atc/sql/schema.py
mrmasterplan: no, but this is outdated now.

src/atc/sql/schema.py
mrmasterplan: yes, but this is too challenging right now. it needs some upstream support for token positions. I will not wait for that right now.

src/atc/sql/schema.py
mrmasterplan: it inverts the boolean. But no longer relevant.

src/atc/schema_manager/schema_manager.py
mrmasterplan: Indeed it will fix that. It closes #219

mrmasterplan:
Hi @GustavNK, i think you also encountered issue #219. This is the fix for you. All tests pass now.

GustavNK (approved):
Looks good

PR #227, opened 2022-12-19
feat: use centralized constants

LauJohansson:


JeppeBlixen (commented):
Maybe we should use Enums instead.
Something like this as an example for table_property:

from enum import Enum

 class TABLE_PROPERTY_NAME(Enum):
    PATH = "path"
    NAME = "name"

def table_property(
        self, table_id: str, property_name: TABLE_PROPERTY_NAME, default_value: str = None
    ):
    property_value = self.get_all_details().get(
            f"{table_id}_{property_name.value}", default_value
        )

table_property(property_name=TABLE_PROPERTY_NAME.PATH)

LauJohansson:
> Maybe we should use Enums instead. Something like this as an example for table_property:
> 
> from enum import Enum
> 
> class TABLE_PROPERTY_NAME(Enum): PATH = "path" NAME = "name"
> 
> def table_property( self, table_id: str, property_name: TABLE_PROPERTY_NAME, default_value: str = None ): property_value = self.get_all_details().get( f"{table_id}_{property_name.value}", default_value )
> 
> table_property(property_name=TABLE_PROPERTY_NAME.PATH)

Good point. I will try using StrEnum.

PR #228, opened 2022-12-20
Feature/tab lock write sql

LauJohansson:
To improve the write opreration to SQL server the SqlServer class should use TabLock.

tableLock= TRUE 
*Implements an insert with TABLOCK option to improve write performance*
https://learn.microsoft.com/en-us/sql/connect/spark/connector?view=sql-server-ver16#supported-options

Microsoft has also done performance comparisons of the mssql connector using TABLOCK. See: 
https://github.com/microsoft/sql-spark-connector/blob/master/README.md#performance-comparison

See also:
https://learn.microsoft.com/en-us/sql/connect/jdbc/using-bulk-copy-with-the-jdbc-driver?view=sql-server-ver16

### DIFF
--- a/tests/cluster/sql/test_deliverysql.py
+++ b/tests/cluster/sql/test_deliverysql.py
@@ -309,6 +309,19 @@ class DeliverySqlServerTests(DataframeTestCase):
             expected_data=expectedData,
         )
 
+    def test20_big_data_set(self):
+        # Ensure table creation
+        self.sql_server.drop_table(self.table_name)
+        self.create_test_table()
+
+        # Create test data
+        df = self.create_data()
+
+        # Write as big data set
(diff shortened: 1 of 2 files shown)


JeppeBlixen (approved):

tests/cluster/sql/test_deliverysql.py
```diff
@@ -309,6 +309,19 @@ def test19_upsert_to_table_overwrite_empty_target(self):
             expected_data=expectedData,
         )
 
+    def test20_big_data_set(self):
```
JeppeBlixen, line 312: This test has not directly anything to do with the table lock option? It was just missing right?

LauJohansson: Correct. It was just a missing test!

PR #229, opened 2023-01-03
Feature/batch incremental extractor

LauJohansson:
**Introduction**
When working with batch processing - an incremental extractor is needed.

This extractor is especially helpful when doing upserts to a target table like this ETL process:

*Use the IncrementalExtractor for incrementally read of data*
*Transform your data*
*Use UpsertLoader for upserting data to target table*

**Why not use Autoloader?**
Autoloader is meant for reading data in a Structured Streaming setup. Batch processing then still needs a incremental load class.


src/atc/etl/extractors/incremental_extractor.py
```diff
@@ -0,0 +1,52 @@
...
+    NB: It is not recommended to use this on Eventhub data.
+        Use EventHubCaptureExtractor instead.
+
+    """
+
+    def __init__(
+        self,
+        handle_source: Readable,
+        handle_target: Readable,
+        time_col: str,
```
LauJohansson, line 22: the time col could be different between source and target

LauJohansson: implemented

JeppeBlixen (approved):

PR #255, opened 2023-02-17
feat: make it possible to save bodyjson as string

LauJohansson:
It should be possible to keep the json body as it is.


LauJohansson:
@mrmasterplan - maybe we could pair program and make some tests of the transformer?

mrmasterplan (approved):

Issue #254, opened 2023-02-17
Make it possible to save eventhub body as STRING


LauJohansson:
Using Multi-hop/Medallion architecture - a bronze table should.

*"The Bronze layer is where we land all the data from external source systems. The table structures in this layer correspond to the source system table structures "as-is," along with any additional metadata columns that capture the load date/time, process ID, etc. The focus in this layer is quick Change Data Capture and the ability to provide an historical archive of source (cold storage), data lineage, auditability, reprocessing if needed without rereading the data from the source system."*

https://www.databricks.com/glossary/medallion-architecture

It should be possible to keep the json body as it is.

mrmasterplan:
This sounds very reasonable actually. Let's discuss.

PR #258, opened 2023-03-01
Eventhub data to bronze "as-is". Unpack to silver

LauJohansson:
This is another kind of solution than #254.

Instead of altering the `EhJsonToDeltaOrchestrator` (which is used in production now), here is a suggestion:

- Class EhToDeltaBronzeOrchestrator
  - Does almost the same as `EhJsonToDeltaOrchestrator`. But does not unpack the data. The data stays as-is in bronze.
  - Always _appends_ new eventhub data to bronze
- Class EhToDeltaSilverOrchestrator
  - Extracts data from bronze
  - Unpacks the data using the already implemented `EhJsonToDeltaTransformer` 
  - Load modes available: upsert, append and overwrite


src/atc/orchestrators/eh2bronze/EhToDeltaBronzeOrchestrator.py
```diff
@@ -0,0 +1,41 @@
+from atc.delta import DeltaHandle
```
LauJohansson, line 1: needs documentation

src/atc/orchestrators/eh2bronze/EhToDeltaBronzeOrchestrator.py
```diff
@@ -0,0 +1,41 @@
...
+from atc.eh.EventHubCaptureExtractor import EventHubCaptureExtractor
```
LauJohansson, line 2: needs unittesting?

src/atc/orchestrators/eh2silver/EhToDeltaSilverOrchestrator.py
```diff
@@ -0,0 +1,69 @@
+from atc.delta import DeltaHandle
```
LauJohansson, line 1: needs documentation

src/atc/orchestrators/eh2silver/EhToDeltaSilverOrchestrator.py
```diff
@@ -0,0 +1,69 @@
...
+from atc.etl import EtlBase, Orchestrator
```
LauJohansson, line 2: needs unittesting - test the two if-else statements

docs/orchestrators/README.md
```diff
@@ -31,3 +31,101 @@ There may be cases where only a subset of rows is desired to be extracted in the
...
+
+*"The Bronze layer is where we land all the data from external source systems. The table structures in this layer correspond to the source system table structures "as-is," along with any additional metadata columns that capture the load date/time, process ID, etc. The focus in this layer is quick Change Data Capture and the ability to provide an historical archive of source (cold storage), data lineage, auditability, reprocessing if needed without rereading the data from the source system."*
+
+[Source: Databricks.com](https://www.databricks.com/glossary/medallion-architecture)
+
+Using the orchestrator without adding any filtering `.filter_with` the output schema is the following:
+
+| **Column Name**   | **Data type** | **Explanation**                                                                                                    |
+|-------------------|---------------|--------------------------------------------------------------------------------------------------------------------|
+| BodyId            | Long          | A ID generated to give a unique id for each row in the bronze table.  Calculated based on sha2 hashing the *Body*. |
```
mrmasterplan, line 57: Can you please hash the Body *and* the EnueuedTimestamp. It is absolutely allowed to resend the same message multiple times. The enqueued timestamp will usually differ in those cases.

This does not even solve all use cases, because it is possible to bulk-submit to EventHub where the Body and the EnueuedTimestamp of multiple messages can both batch if they were submitted in the same bulk. But i think that it is OK to disregard this case since I cannot think of a legitimate use case where one might do that.

src/atc/orchestrators/eh2bronze/EhToDeltaBronzeTransformer.py
```diff
@@ -0,0 +1,73 @@
...
+
+from atc.etl import Transformer
+from atc.tables import TableHandle
+
+
+class EhToDeltaBronzeTransformer(Transformer):
+    """
+    This class transforms eventhub data into the following schema:
+
+    | BodyId | Body | EnqueuedTimestamp | StreamingTime | pdate |
```
mrmasterplan, line 13: you have to keep the properties column also. We have seen cases where e.g. the message type was added exclusively in the properties column

src/atc/orchestrators/eh2bronze/EhToDeltaBronzeTransformer.py
```diff
@@ -0,0 +1,73 @@
...
+        ]
+
+    def process(self, df: DataFrame) -> DataFrame:
+        target_df = self.target_dh.read()
+
+        assert set(self._eh_cols).issubset(target_df.columns)
+
+        # Generate id for the eventhub rows using hashed body
+        df = df.withColumn(
+            "BodyId",
```
mrmasterplan, line 45: as said above, please include the Enqueued Timestamp, and maybe even also the properties column to create this hash.

docs/orchestrators/README.md
```diff
@@ -31,3 +31,101 @@ There may be cases where only a subset of rows is desired to be extracted in the
 process. Here the orchestrator offers the method `.filter_with` which allows 
 additional transformation steps to be injected before the rows are finally appended 
 to the delta table.
+
+# Eventhub to medallion architecture
```
mrmasterplan, line 35: Very nice documentation!!

mrmasterplan (changes requested):

docs/orchestrators/README.md
LauJohansson: Added a new hash column: "EventhubRowId"

src/atc/orchestrators/eh2bronze/EhToDeltaBronzeTransformer.py
LauJohansson: I have now included the full eventhub schema

mrmasterplan (approved):
This is a solid piece of work! I expect that we will have ideas for changes and improvements as soon as we start using it, but for now we probably can make se biggest improvement to the code by actually using it. Approved for merge!

LauJohansson:
Suggestions from @RadekBuczkowski :

Documentation: Add, that bodyid is the way to compare streaming and batch

Consider removing sequencenumber (gives not necessarily any value to the ETL)
Consider removing offset (gives not necessarily any value to the ETL)

In the orchestrators, use an enum instead of string (mode="upsert").
When using append - why not use incremental extraction too?

PR #260, opened 2023-03-07
feat: task entry point and module helper, for automatically discovery of entry points

MartinBoge:
This PR brings the TaskEntryPoint base class. When inheriting from this class the "task" abstract method is implemented, thus allowing for automatically discovering entry points.


src/atc/entry_points/task_entry_point.py
```diff
@@ -0,0 +1,8 @@
+from abc import ABC, abstractmethod
+
+
+class TaskEntryPoint(ABC):
```
LauJohansson, line 4: add docstring

docs/entry_points/README.md
```diff
@@ -0,0 +1,5 @@
+# Entry points
+
+The `TaskEntryPoint` is a fundamental class that can be used when developing a Python module to be executed via an entry point in a Python wheel.
+
+When inheriting from this class we ensure the implementation of the `task` method such that these can be easily discovered and manipulated throughout the package.
```
LauJohansson, line 5: add an example of how to use it / how it would look like when inherit from it

MartinBoge: Added

LauJohansson (approved):

PR #263, opened 2023-03-10
Fix/increm extract overlap period

lasseaj:
This PR will fix so overlapping periods in incremental extractors  works correctly when the target table is empty.


LauJohansson (commented):
The bug in the tests is not one you have made.

The CLI has been updated. So either fix the issue in tests.local.configurator.test_configurator_cli.TestConfiguratorCli or maybe rebase/merge the newest changes to your branch.

PR #264, opened 2023-03-10
use test job submission from atc-dataplatform-tools

mrmasterplan:


tests/cluster/job1/delta/est_sparkexecutor.py
```diff
@@ -3,8 +3,8 @@
 from atc import Configurator
 from atc.delta import DbHandle, DeltaHandle
 from atc.spark import Spark
-from tests.cluster.delta import extras
-from tests.cluster.delta.SparkExecutor import SparkSqlExecutor
```
LauJohansson: this file should be renamed to "test_sparkexecutor.py".

PR #268, opened 2023-03-24
Instead of casting exception "The selected slice returned no data" return empty data frame

RadekBuczkowski:
Instead of casting exception "The selected slice returned no data" return an empty dataframe from eventhub if data slice returns nothing.


LauJohansson (approved):

LauJohansson:
There is no test of the performance... reading the schema from the path could might be slow? I have still approved.

